PR #78, opened 2013-07-31
Fix config.assets.cache_store not working anymore

ericboehs:
In Rails 3.2 setting `config.assets.cache_store` would allow changing the cache store sprockets used. This feature was removed when sprockets was ported to a separate repository (for reasons unknown to me). I've added it back as well as added tests for it.

I had to `bundle exec rake test` to get the tests to pick up my changes to the railtie.

For environments without the option of a shared tmp directory, this speeds up asset compilation by 4 times in some scenarios (from 66 seconds to 16). Or on my local machine, 16 seconds to 5 (if you subtract out the 4.5 seconds it takes rails to boot on my local, that's a significant improvement: 12 seconds down to <1).  

**Edit:** By shared tmp directory I mean a persisted tmp directory across deploys (e.g. Heroku doesn't have a persisted tmp directory).

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -22,8 +22,14 @@ module Rails
       @assets ||= Sprockets::Environment.new(root.to_s) do |env|
         env.version = ::Rails.env
 
-        path = "#{config.root}/tmp/cache/assets/#{::Rails.env}"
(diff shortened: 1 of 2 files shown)


viniciusnz:
+1 @ericboehs , just used a slightly different version of your code to use file lookups on [this issue](Compass/compass-rails#98), thanks a lot!

PR #109, opened 2014-01-28
Change README for avoid misleading related to vendor/assets support

xdite:
https://github.com/rails/rails/pull/7968  from this ticket. Default design was already changed a years ago. Should change README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Set to `true` to enable additional runtime error checking. Recommended in the `d
 
 **`config.assets.paths`**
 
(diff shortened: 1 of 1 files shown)


prusswan:
> No. css and javascript files in lib/assets and vendor/assets are still valid. We are only not including non-css and non-javascript files by default.

@rafaelfranca Is this still accurate as of Rails 4.2? If so, sprockets appear to have inconsistent behaviour on different OS - on Windows (tested with 7 and 10), `= require_tree .` fails to include/process the js/css files in `vendor/assets`. It works out of the box on Ubuntu, however.

PR #215, opened 2015-01-20
Import append_assets_path initializer

josh:
Alright, @rafaelfranca

Heres the idea for moving the `append_assets_path` into this plugin.

Then in rails/master we could completely remove `config.assets` and this guy. This code would still cover people running Rails 4.2. And eventually we can remove the guard when we only need to support Rails 5.

Related discussion

* https://github.com/rails/rails/pull/18607

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -40,7 +40,7 @@ module Rails
   class Engine < Railtie
     # Skip defining append_assets_path on Rails <= 4.2
     unless initializers.find { |init| init.name == :append_assets_path }
-      initializer :append_assets_path, group: :all do |app|
+      initializer :append_assets_path, :group => :all do |app|
         app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)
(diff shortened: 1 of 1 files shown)


lib/sprockets/railtie.rb
```diff
@@ -36,6 +36,17 @@ def assets
     # Returns Sprockets::Manifest for app config.
     attr_accessor :assets_manifest
   end
+
+  class Engine < Railtie
```
josh: Is extending initializers on `Engine` allowed?

rafaelfranca: Yes, it is, although it is not the best way of doing it.

Since we want to detect if the initializer was already defined I think we can't do better.

When we drop support to Rails 4 we can change it to instead of reopening `Rails::Engine` just define a `Sprockets::Rails::Engine` instead of a `Railtie`.

lib/sprockets/railtie.rb
```diff
@@ -36,6 +36,23 @@ def assets
...
+    # Skip defining append_assets_path on Rails <= 4.2
+    unless initializers.find { |init| init.name == :append_assets_path }
+      initializer :append_assets_path, :group => :all do |app|
+        if paths["app/assets"].respond_to?(:existent_directories)
```
rafaelfranca: Why bother with versions < 4 if the initializer is defined and we will not enter this code? It is just because Rails 3.0?

josh: :/ yeah, I just wanted to make the tests pass. This is merging into the old 2.x. I'll drop this condition when pulling this patch into master.

lib/sprockets/railtie.rb
josh: > When we drop support to Rails 4 we can change it to instead of reopening Rails::Engine just define a Sprockets::Rails::Engine instead of a Railtie.

Ah, cool, I didn't realize that would work.

PR #221, opened 2015-02-03
Enable digest by default

josh:
Now the default for dev and prod mode in Rails initial templates.

Change going into 3.x line.


rafaelfranca:
We are already doing at the [generated environments](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/config/environments/development.rb.tt#L37) so flipping the default is not a problem for 3.x. :shipit:

josh:
@rafaelfranca thanks, thats what I meant to say.

PR #222, opened 2015-02-03
Always initialize assets environment in rake task

josh:
#220 causes a slight regression but brings up this weird issue again.

If `config.assets.compile = false` in `production.rb` than you can't even do `rake assets:precompile`.

This is a little work around to force the environment to build even in this case.

@rafaelfranca the environment weirdness between `rake assets:precompile` has always bugged me. Given 3.x can have backwards incompatible changes, have any idea for directions we can move into to make this more clear?

I really wish `rake assets:precompile` would work no matter what and always minified the assets even in RAILS_ENV=development. Would like to make it easier for people who compile assets in an isolate environment and upload them to a CDN. This sometimes happens locally on there machine and othertimes in production like environments that might not have a db.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -69,7 +69,8 @@ module Sprockets
       Sprockets::Rails::Task.new(app)
     end
 
-    def build_environment(app, initialized: app.initialized?)
+    def build_environment(app, initialized = nil)
+      initialized = app.initialized? if initialized.nil?
       unless initialized
         ::Rails.logger.warn "Application uninitialized: Try calling YourApp::Application.initialize!"
       end
@@ -130,7 +131,7 @@ module Sprockets
       config = app.config
 
       if config.assets.compile
-        app.assets = self.build_environment(app, initialized: true)
+        app.assets = self.build_environment(app, true)
         app.routes.prepend do
           mount app.assets => config.assets.prefix
         end
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -69,7 +69,11 @@ module Sprockets
       Sprockets::Rails::Task.new(app)
     end
 
-    def self.build_environment(app)
+    def build_environment(app, initialized: app.initialized?)
+      unless initialized
+        ::Rails.logger.warn "Application uninitialized: Try calling YourApp::Application.initialize!"
+      end
+
       config = app.config
       env = Sprockets::Environment.new(app.root.to_s)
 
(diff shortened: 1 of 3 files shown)


rafaelfranca:
> I really wish rake assets:precompile would work no matter what and always minified the assets even in RAILS_ENV=development.

Yeah, I'd like too but there are some configuration that still changes between environment like the assets prefix and the assets host. I don't have any idea on how change this to not use environment information.

josh:
Yeah, I dunno either :/

I think at least this compile direction will at least allow `rake assets:precompile` to do *something*.

The other issue I'm still worried about is building against an uninitialized app. Is there any flag thats set after `initialize!` is called? It'd be nice to check `app.initialized?` (maybe that already works) then maybe print a warning saying, hey, you attempted to compile assets against an non-initialized app, some settings might be wrong, etc.

rafaelfranca:
> It'd be nice to check app.initialized? (maybe that already works) then maybe print a warning saying, hey, you attempted to compile assets against an non-initialized app, some settings might be wrong, etc.

Yeah, `initialized?` already exists, and a warning would be great.

josh:
> Yeah, initialized? already exists, and a warning would be great.

Bah, so `app.initialized?` does work but `app.initialized? == false` inside the `config.after_initialize` block.

lib/sprockets/railtie.rb
```diff
@@ -69,7 +69,11 @@ def configure(&block)
       Sprockets::Rails::Task.new(app)
     end
 
-    def self.build_environment(app)
+    def build_environment(app, initialized: app.initialized?)
```
josh: A little hacky, just allowing the after_init callback to skip the warning.

lib/sprockets/railtie.rb
```diff
@@ -69,7 +69,11 @@ def configure(&block)
...
+      unless initialized
+        ::Rails.logger.warn "Application uninitialized: Try calling YourApp::Application.initialize!"
```
josh: @rafaelfranca does this error make sense?

Also, I was wondering why `app.logger` doesn't exist.

rafaelfranca: Not sure, it seems we are talking about an internal thing of Rails. Users never had to worry about initializing an application. Do you know when the application is not initialized at this point?

> Also, I was wondering why app.logger doesn't exist.

The application doesn't have logger. It is a global thing in the Rails module.

josh: > Do you know when the application is not initialized at this point?

If you override the `assets:environment` task, which is permitted, you could forget or screw up initializing the app on your own which could lead to an uninitialized app instance here. Yeah, this won't affect people on the golden path.

rafaelfranca: I see. So I think the message is good.

PR #224, opened 2015-02-08
Experimental SRI support

josh:
Sprockets 3.x builds in some plumbing for getting [SRI](http://www.w3.org/TR/SRI/) hashes. This next step exposes the API at the Rails asset tag helper level.

``` ruby
javascript_include_tag :application, integrity: true
# => "<script src="/assets/application.js" integrity="ni:///sha-256;TvVUHzSfftWg1rcfL6TIJ0XKEGrgLyEq6lEpcmrG9qs?ct=application/javascript"></script>"
```

* The `integrity` format is still a little in flux. The format is generated in Sprockets itself, so spec changes *may* require core changes but nothing at the view level. The `integrity` attribute itself seems pretty established.
* `integrity: true` only works on Sprockets assets, not on `public/` static ones. It'd be cool if at some point in the future, Rails bakes this in directly and provides an asset integrity hook similar to the current `compute_asset_path` hook.
* Base branch targets `master` which is currently released as 3.x beta gems. This isn't going into 2.x stable.

Eventually this maybe something we can enable by default for Rails apps with a global config

``` ruby
config.assets.sri = true
```

/cc @rafaelfranca @jeremy @mastahyeti @mikewest


rafaelfranca:
I liked the idea. :shipit: from my side

PR #228, opened 2015-03-17
Don't precompile assets without an extension

elia:
Sprockets 2.x dies with anything without an extension, 
READMEs from vendored assets are the most frequent offenders.

Fixes https://github.com/sstephenson/sprockets/issues/347.

I know the issue has been closed but the fix is so easy and simple I felt it was worth retrying.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -58,7 +58,8 @@ end
 module Sprockets
   class Railtie < ::Rails::Railtie
     LOOSE_APP_ASSETS = lambda do |filename, path|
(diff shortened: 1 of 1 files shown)


rafaelfranca:
@elia correct if I'm wrong but I think we only precompile all the files that doesn't have `.js` or `.css` extension in the `app/assets` folder. If it is precompiling READMEs from vendored assets does it means that your vendored assets are inside the `app` folder?

elia:
Checked again the code and apparently there was another lambda in the app that was patched along with the introduction of this fix.

Sorry for the time waste, I'll try to reproduce the error and reopen if needed.

PR #231, opened 2015-03-24
Update the test suite to account for named pipelines and the manifest name change.

lucasmazza:
These fixes aren't 100% yet, since some tests are failing due mismatched digests, even though we read the digest directly from the asset itself - I'm gonna take a deeper look on this later.

/cc @rafaelfranca @josh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ sudo: false
 rvm:
   - 1.9.3
   - 2.0.0
-  - 2.1
-  - 2.2
+  - 2.1.5
+  - 2.2.1
 
 gemfile:
   - Gemfile
@@ -28,7 +28,7 @@ matrix:
     - gemfile: test/gemfiles/Gemfile.rails-5.0.x
       rvm: 2.0.0
     - gemfile: test/gemfiles/Gemfile.rails-5.0.x
-      rvm: 2.1
+      rvm: 2.1.5
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
     - gemfile: Gemfile
       rvm: 2.0.0
     - gemfile: Gemfile
-      rvm: 2.1
+      rvm: 2.1.5
     - gemfile: test/gemfiles/Gemfile.rails-5.0.x
       rvm: 1.9.3
     - gemfile: test/gemfiles/Gemfile.rails-5.0.x
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -428,35 +428,35 @@ class DebugHelperTest < NoHostHelperTest
   def test_javascript_include_tag
     super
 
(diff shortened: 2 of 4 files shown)


josh:
Nice.

Right now `?body=1` is just aliased to use the body pipeline. https://github.com/rails/sprockets/blob/master/lib/sprockets/server.rb#L47

Just curious if you seeing any issues with how it works now. I was hoping this would be backwards compatible, but it'd be great to modernize this as well.

lucasmazza:
@josh I'm ok with how it works now, but I wonder if both `self` and `?body=1` are doing the same thing we could remove one of those from the code.

The current failures are related to some assets were their digest are changing between the test setup and when the helper generates the `script`/`link` tags, which shouldn't happen - no luck tracking this down so far.

josh:
>  I wonder if both self and ?body=1 are doing the same thing we could remove one of those from the code.

Yeah, that sounds good. `foo.self.js` would be preferred in that case. Maybe we can remove the `?body=1` thing from sprockets 4.x if everything gets moved over.

lucasmazza:
@josh would you mind taking a look on these failures later? From what I've found so far, some assets are ending with different digest values because at one point the digest is calculated with the concatenated source of another asset that was required beforehand, but I couldn't find out why we had the different source when processing the digest.

josh:
Yeah, I'll to pull this down tonight or tomorrow.

Technically `"/assets/foo-#{@foo_css_digest}.css?body=1"` is actually a bug. Theres an edge case for compatible that basically discards the digest if `?body=1` is set.

I think `@foo_css_digest` is set to `env["foo.css"].digest_path` when it should actually be `env["foo.self.css"].digest_path` or `env.find_asset("foo.css", pipeline: :self).digest_path`.

josh:
Made a reverse PR to your fork, @lucasmazza. See https://github.com/lucasmazza/sprockets-rails/pull/1.

lucasmazza:
@josh awesome, the build is green now.

On a side note, the `Gemfile` and `Gemfile.rails-5.0.x` are both pointing to rails master. We revert the main Gemfile to use a released version and let the test gemfiles target the rails branches.

.travis.yml
```diff
@@ -5,8 +5,8 @@ sudo: false
 rvm:
   - 1.9.3
   - 2.0.0
-  - 2.1
-  - 2.2
+  - 2.1.5
```
rafaelfranca: Why specify the version? 2.1 is valid and will test against the latest release of 2-1 branch of Ruby.

rafaelfranca:
Awesome

.travis.yml
lucasmazza: `2.2` was loading `2.2.0` (which is broken with Rails 5), so I had to expand it. I decided to do the same with the `2.1` to maintain the list format consistency.

rafaelfranca: :+1:

PR #232, opened 2015-03-24
Add `app/assets` to the load path.

lucasmazza:
Now we can compile manifests located at `app/assets/manifest.js` or something like it.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -33,6 +33,7 @@ module Rails
       initializer :append_assets_path, :group => :all do |app|
         app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
+        app.config.assets.paths.unshift(*paths["app"].existent_directories.grep(/\/assets\z/))
         app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)
       end
     end


lib/sprockets/railtie.rb
```diff
@@ -33,6 +33,7 @@ class Engine < Railtie
       initializer :append_assets_path, :group => :all do |app|
         app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
+        app.config.assets.paths.unshift(Rails.root.join("app/assets").to_s)
```
rafaelfranca: This will get just the application app/assets, but we may want the engine's path.

lucasmazza: right, how we can do that? I couldn't get the `paths['app/assets']` as String here.

rafaelfranca: I think only getting `paths['app']` and filtering just `app/assets`

lucasmazza: Something like `paths['app'].existent_directories.grep(/\/assets$/)` would do then :+1:

josh:
So originally I thought this was going to be a terrible idea since it creates overlapping load paths. But I ended up doing the same thing in github.com to make the manifest.js thing work as well. Its been working fine so far, I was just hesitant to upstream it. But, lets do it :+1: I'm sure other people will help find the weird issues that come up.

josh:
@lucasmazza thanks again for pushing the manifest stuff forward. Really curious to hear what you think of the approach after you've got it all setup.

lucasmazza:
@josh thanks for all the work on Sprockets so far. I think the manifest is an upgrade to the `compile` setting we have and can be a more expressive way to configure this part of the app. My current project is fairly simple to work as a benchmark for the upgrade but it was smooth so far. 

I'm curious if would make sense to have these manifests written in Ruby instead of comments in the assets, as they probably won't produce any output directly - a `manifest.js` would just link other real assets that need to be compiled and become an empty file when the precompile task is invoked.

josh:
> I'm curious if would make sense to have these manifests written in Ruby instead of comments in the assets, as they probably won't produce any output directly - a manifest.js would just link other real assets that need to be compiled and become an empty file when the precompile task is invoked.

Don't disagree. The current suggestion is mostly an accident feature that works for free. I was think it could be some special `manifest.yml` file or something. But a fully scriptable ruby file could be interesting too. Maybe theres some ways to shoot yourself in the foot there. Technically this works today if you want to experiment.

``` erb
<%= link_asset "application.js" %>
<%= link_asset "application.css" %>
```

But yeah, mostly been waiting on feedback for how people thing it should work. So :heart: for trying to work though it. I'd definitely would love to move some of this forward more.

josh:
Do we just need to merge in the latest master to get this build green?

lucasmazza:
:+1: rebased with master

lucasmazza:
I guess I've overlooked this a bit, but this code isn't executed for rails 4 apps. Should we patch this someplace else to get the same behaviour for Rails 4 and 5?

rafaelfranca:
I don't think we changed the paths code in master. I'm pretty sure 4.2 and 5 are using the same code for `Paths` class.

lucasmazza:
@rafaelfranca the initializer is skipped if it is already defined, and that's the case in 4.x apps https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L31-L32

rafaelfranca:
Ah, right. I was thinking that the new path didn't worked, but forgot that we had this code inside of rails before

dhh:
This change slowed our test suite by 34%. And that's the whole suite including model tests etc. So the impact in integration tests much be massive.

jeremy:
I'd like to use a separate dir for the manifest. Using `app/assets/config/manifest.js` currently. Adding `app/assets` itself confuses ppl since it suggests you can/should refer to `stylesheets/application` instead of just `application`.

This is the one asset I'd like to refer to outside the load paths, even. I'm liking the manifest concept overall. Plugin/engine integration needs work still. The `config.assets.precompile += …` "DSL" still works great for them. Wonder whether our manifest should be a Ruby DSL that constructs an internal asset with our link deps.

jeremy:
(Re. speed: This adds tons of paths, one for each engine, which means lots more file-digest deps. If you reference a lot of assets, you'll end up with thousands of file-digest deps, scaling with # of assets * # of paths they could appear in.)

rafaelfranca:
:+1: for a separate dir.

PR #238, opened 2015-04-22
Move assets route internal decoration.

arthurnn:
We can't depend on assets_prefix outside sprockets.

With this, we can create a rails --api without sprockets and `rake routes` wont fail.

After this is merged, we need to patch Rails. I will send a PR there too.

review @spastorino @rafaelfranca 
@zzak paired with me (he did most of the code :smiley_cat:)

### DIFF
--- a/lib/sprockets/rails/route_wrapper.rb
+++ b/lib/sprockets/rails/route_wrapper.rb
@@ -1,0 +1,19 @@
+module Sprockets
+  module Rails
+    module RouteWrapper
+      def self.included(klass)
+        klass.class_attribute(:assets_prefix)
+        klass.class_eval do
+          def assets_prefix
+            self.class.assets_prefix
+          end
+
+          def internal_with_sprockets?
+            internal_without_sprockets? || path =~ %r{\A#{assets_prefix}\z}
+          end
+          alias_method_chain :internal?, :sprockets
+        end
+      end
+    end
+  end
+end
--- a/lib/sprockets/rails/route_wrapper.rb
+++ b/lib/sprockets/rails/route_wrapper.rb
@@ -1,15 +1,19 @@
 module Sprockets
   module Rails
     module RouteWrapper
-      def self.included(klass)
-        klass.class_attribute(:assets_prefix)
-        klass.class_eval do
-          def assets_prefix
-            self.class.assets_prefix
-          end
 
+      def internal_assets_path?
+        path =~ %r{\A#{self.class.assets_prefix}\z}
+      end
+
+      def internal?
+        super || internal_assets_path?
+      end
+
+      def self.included(klass)
+        klass.class_eval do
           def internal_with_sprockets?
-            internal_without_sprockets? || path =~ %r{\A#{assets_prefix}\z}
+            internal_without_sprockets? || internal_assets_path?
           end
           alias_method_chain :internal?, :sprockets
         end
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -6,6 +6,7 @@ require 'active_support/core_ext/numeric/bytes'
 require 'sprockets'
 require 'sprockets/rails/context'
 require 'sprockets/rails/helper'
+require 'sprockets/rails/route_wrapper'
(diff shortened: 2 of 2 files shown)


lib/sprockets/rails/route_wrapper.rb
```diff
@@ -0,0 +1,19 @@
...
+        klass.class_attribute(:assets_prefix)
+        klass.class_eval do
+          def assets_prefix
+            self.class.assets_prefix
+          end
+
+          def internal_with_sprockets?
+            internal_without_sprockets? || path =~ %r{\A#{assets_prefix}\z}
+          end
+          alias_method_chain :internal?, :sprockets
```
rafaelfranca: This will give us warning with Rails master. Also I think this entire code should only be applied if we are in Rails master.

rafaelfranca:
LGTM besides the fact that we only need it on Rails master and we can't use `alias_method_chain`

lib/sprockets/rails/route_wrapper.rb
spastorino: Agreed, avoid this :)

josh: Curious what `RouteWrapper#internal?` does in Rails itself?

spastorino:
I don't think this is the right solution. I wouldn't override AV internal methods from another.

Currently sprockets-rails sets AV's config.assets when sprockets-rails is turned on. So if assets is defined in AV we've sprockets running and if it's not defined sprockets was turned off. So the current code that checks for its existence is already right.

Is it neither a good solution from my point of view but I think the existing code is better than the proposed PR. 

What we could do is provide an AV class that handles the asset stuff and then sprockets-rails is turned on it can inherit from that class and provide another that adds what it needs to add.

lib/sprockets/rails/route_wrapper.rb
arthurnn: @josh I am trying to fix https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing/inspector.rb#L48 , which on action_dispatch has a reference to sprockets.
On rails api when we dont require sprockets, that it is not working.

josh: Ah, that makes sense. Thanks for the context!

arthurnn:
@rafaelfranca @chancancode I added the `prepend` for when we can use that instead of `alias_method_chain`

me and @spastorino , kinda chat about this. This solution is not pretty, What we should have is some API on ActionView and sprockets could set those. However that will require a bigger refactor. For now, this fixes the problem, and we can discuss a better API btwn actionview and sprockets in the future.

lib/sprockets/rails/route_wrapper.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      def internal?
+        super || internal_assets_path?
+      end
+
+      def self.included(klass)
+        klass.class_eval do
+          def internal_with_sprockets?
+            internal_without_sprockets? || internal_assets_path?
```
rafaelfranca: Will this work? I think when you include you will add this `internal?` implementation that calls super.

arthurnn: Initially I thought that too, and I tested and it worked. After investigating I figured out why.
When we include this module, the class already defines the `internal?` method, so the `internal?` implementation in the module will never been called, as ruby calls the class method first, and only the alias_method_chain version will work.

zzak:
Just to re-iterate, the point of this patch is to allow us to remove Sprockets dependent code from action_dispatch's `RouteWrapper` class.

We can make a separate patch/discussion to fix the existing monkey patches in sprockets-rails.

spastorino:
:heart: :heart: :heart:, now fix it in Rails :)

arthurnn:
la haré! =)

arthurnn:
@spastorino https://github.com/rails/rails/commit/757a2bc3e3e52a5d9418656928db993db42b741b 

You should rebase your rails-api branch, and remove the changes around that.

spastorino:
thanks!

PR #241, opened 2015-04-28
Prevent alias logical paths passed to asset_path

josh:
Theres a few other edge cases that will slip by the `AssetNotPrecompiled` check that are both issues today on 2.x and 3.x.

* `index` assets can **only** be referenced by their shorthand alias and not by the full path
* bower paths must be referenced by their full path and never there shorthand

The index case is annoying and I'm looking to address it in https://github.com/rails/sprockets/pull/36.

Though, I think this check is probably a good idea no matter what happens in sprockets core.

We basically assert that the input `path` which would be passed directly to `manifest.assets` in production matches the `logical_path` as it would be written out to disk.

The check only affects AV helpers, nothing within asset generation within sprockets.

/cc @arthurnn @rafaelfranca @matthewd

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -15,6 +15,15 @@ module Sprockets
         end
       end
 
+      class AssetAliasUsed < StandardError
+        def initialize(actual, expected)
+          msg = "Asset was linked to from an alias rather than its exact path. " +
+                "Alias resolving may not be available in production.\n" +
+                "Use #{expected.inspect} instead of #{actual.inspect}"
+          super(msg)
+        end
+      end
+
       include ActionView::Helpers::AssetUrlHelper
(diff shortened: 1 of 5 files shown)


rafaelfranca:
So now when a bower path is referenced by its shorthand we will raise an exception?

josh:
> So now when a bower path is referenced by its shorthand we will raise an exception?

From a view `javascript_tag` helper, yes.

Its unlikely anyone is using that since it doesn't work if you deployed it to production.

rafaelfranca:
Got it. :shipit:

josh:
Rad.

wdyt about backporting this to sprockets-rails 2.x? Think its worth it?

rafaelfranca:
> wdyt about backporting this to sprockets-rails 2.x? Think its worth it?

I think so. At least it will avoid some surprising erros on production.

PR #248, opened 2015-05-12
Fix 2.x build

josh:
Revise tests according to behavior changes in Sprockets 3.1 around index specialness.

/cc @rafaelfranca


josh:
Bah, theres still some `TestTask#test_clean_with_keep_specified [/home/travis/build/rails/sprockets-rails/test/test_task.rb:138]` fail on old versions. Not sure whats up.

PR #251, opened 2015-05-14
Revert asset alias check

josh:
@rafaelfranca wdyt about just rolling this back? It should fix @dhh's problem in https://github.com/rails/sprockets/pull/36.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -15,15 +15,6 @@ module Sprockets
         end
       end
 
-      class AssetAliasUsed < StandardError
-        def initialize(actual, expected)
-          msg = "Asset was linked to from an alias rather than its exact path. " +
-                "Alias resolving may not be available in production.\n" +
-                "Use #{expected.inspect} instead of #{actual.inspect}"
-          super(msg)
-        end
-      end
-
       include ActionView::Helpers::AssetUrlHelper
       include ActionView::Helpers::AssetTagHelper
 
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Yeah, seems better to revert and come with a solution that will work for that case.

dhh:
Awesome. That does the trick for me.

I can open another ticket for this, but we also hit a big performance regression upgrading. These changes:

`
 GIT
   remote: git://github.com/rails/sprockets-rails.git
-  revision: 0159123f01a54170f651934e1fef9b3f61341fcc
+  revision: 4844424fe71944350cb0645c2d81ac49ae6d3f36
   specs:
     sprockets-rails (3.0.0.beta1)
       actionpack (>= 4.0)
       activesupport (>= 4.0)
-      sprockets (>= 3.0.0.beta, < 4.0)
+      sprockets (>= 3.0.0, < 4.0)
 
 GIT
   remote: git://github.com/rails/sprockets.git
-  revision: 92eb3072790cef9f1e28c9670abeddcf45de8099
-  ref: 92eb3072790cef9f1e28c9670abeddcf45de8099
+  revision: 4e555b82cf43c0ddb0dbeda7c68bc306bc094533
   specs:
-    sprockets (3.0.0.beta.8)
+    sprockets (3.1.0)
       rack (~> 1.0)
`

Take our total suite run time from 54s to 72s on my machine.

josh:
@dhh ah nice.

I had an idea for that but haven't got around to implementing it yet. I'll take another stab on the weekend.

dhh:
Let me know if there's anything else for me to try re: the performance hit.

Issue #234, opened 2015-04-06
Adding app/ to the asset path is a large performance regression


dhh:
Pull request #232 caused a 34% performance hit in our test suite.

rafaelfranca:
Wow! This is huge. I'll take a look

josh:
Woah, yeah, would love to see why. I know an extra search path will cause a little more searching, but not that much.

Your sure its isolated to 21c9c384cd59bf9774f4db7a67d6ec6645e3dfb4? No other changes?

Curious about your integration test suite. Are you doing any parallelization? One thing we make sure we do on the github test suite is warm the asset cache on the prefork hook. Basically just `Rails.application.assets["application.js"]` This ensures we only build it once and theres no chance of separate workers racing to recompute the same data.

dhh:
Just that change. Verified it this morning. Not doing any parallelization
at the moment.

josh:
Can you post your current asset load path set.

```
>> puts Rails.application.assets.paths
```

Wondering if anything else is wrong there.

dhh:
Here's what I get on the version before we expanded the load path:

```
irb(main):001:0> puts Rails.application.assets.paths
/Users/david/37s/apps/bc3/app/assets/fonts
/Users/david/37s/apps/bc3/app/assets/images
/Users/david/37s/apps/bc3/app/assets/javascripts
/Users/david/37s/apps/bc3/app/assets/sounds
/Users/david/37s/apps/bc3/app/assets/stylesheets
/Users/david/37s/apps/bc3/lib/assets/javascripts
/Users/david/37s/apps/bc3/lib/assets/stylesheets
/Users/david/37s/apps/bc3/vendor/assets/javascripts
/Users/david/37s/apps/bc3/vendor/assets/stylesheets
/Users/david/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/bundler/gems/action_cable-bdbbe18f3cc5/lib/assets/javascripts
/Users/david/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/bundler/gems/signal_id-278f61155069/app/assets/images
/Users/david/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/bundler/gems/signal_id-278f61155069/app/assets/javascripts
/Users/david/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/bundler/gems/signal_id-278f61155069/app/assets/stylesheets
[...]

PR #252, opened 2015-05-17
Avoid walking over entire load paths for simple precompile strings

josh:
Related #234.

/cc @rafaelfranca @dhh

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -222,15 +222,23 @@ module Sprockets
           @precompiled_assets ||= begin
             assets = Set.new
 
-            filters = (assets_precompile || []).map { |f|
-              Sprockets::Manifest.compile_match_filter(f)
-            }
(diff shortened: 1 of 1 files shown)


josh:
This stupid "loose asset block" is still going to be slow.

https://github.com/rails/sprockets-rails/blob/6ea7c602bb87e81d011780936cc024d95297ae7c/lib/sprockets/railtie.rb#L46-L49

dhh:
Happy to test our suite against latest sprockets-rails to see what effect it has on the suite run time.

rafaelfranca:
Looks good to me.

Right now I don't see how to remove the loose asset block and keep backward compatibility but maybe we should precompile only assets that are required using some of the existent APIs (like explicit `link` directive and assets helpers) in future versions.

josh:
> Right now I don't see how to remove the loose asset block and keep backward compatibility but maybe we should precompile only assets that are required using some of the existent APIs (like explicit link directive and assets helpers) in future versions.

Yeah, thats the direction I'd like to head in.

Supporting the current Proc API inherently requires iteration over every logical path which is slow and error prone.

PR #253, opened 2015-05-26
Caching in process memory for javascript_include_tag, stylesheet_link_tag

huacnlee:
## Why

Some day I found `javascript_include_tag` and `stylesheet_link_tag` has spend about 2ms in each requests in my project in production environment. 

I saw the `javascript_include_tag` and `stylesheet_link_tag` source code in sprockets-rails it very complex, but that methods it always return same result by same params.

**Why do we caching them? **

In general case, a website not have too much .css, .js links (in production), 
the front-end optimization tell us to [Minimize HTTP Requests](https://developer.yahoo.com/performance/rules.html#num_http), so caches just spend a little memory.

If we caching them, we can reduce 2ms/request.

### Benchmarks:

https://gist.github.com/huacnlee/e942ce9843b5c01bc7dd

```
once call without cache = 2.104
once call cache         = 0.06899999999999999

javascript_include_tag witout cache
                       261.000  i/100ms
javascript_include_tag with cache
                         1.254k i/100ms
javascript_include_tag witout cache
                          3.202k (± 6.1%) i/s -     16.182k
javascript_include_tag with cache
                         49.110k (±10.1%) i/s -    242.022k

Comparison:
javascript_include_tag with cache:    49109.6 i/s
javascript_include_tag witout cache:     3202.2 i/s - 15.34x slower

stylesheet_link_tag witout cache
                       264.000  i/100ms
stylesheet_link_tag with cache
                         1.269k i/100ms
stylesheet_link_tag witout cache
                          3.159k (± 7.9%) i/s -     15.840k
stylesheet_link_tag with cache
                         47.309k (±11.1%) i/s -    233.496k

Comparison:
stylesheet_link_tag with cache:    47308.8 i/s
stylesheet_link_tag witout cache:     3159.1 i/s - 14.98x slower
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,4 @@ gemspec
 
 gem 'actionpack', :github => 'rails/rails', branch: 'master'
 gem 'railties', :github => 'rails/rails', branch: 'master'
+gem 'mocha'
--- a/gemfiles/Gemfile.rails-4.0.x
+++ b/gemfiles/Gemfile.rails-4.0.x
@@ -3,3 +3,4 @@ gemspec :path => ".."
 
 gem "actionpack", "~> 4.0.0"
 gem "railties", "~> 4.0.0"
+gem "mocha"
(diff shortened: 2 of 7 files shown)


u0x01:
结果到现在 Rails Core Team 都没看这个PR

rafaelfranca:
Have we tried to optimize the code path before trying to cache it?

PR #259, opened 2015-06-23
Add support for Sprockets 4.x.

Andreis13:
Now sprockets-rails works with both, sprockets 3.x and sprockets 4.x.
The source maps debug pipeline is available when using sprockets 4.x.

The changes depend on the version of sprockets being specified. (https://github.com/rails/sprockets/pull/64)

### DIFF
--- a/lib/sprockets/rails/context.rb
+++ b/lib/sprockets/rails/context.rb
@@ -17,7 +17,7 @@ module Sprockets
         @dependencies << 'actioncontroller-asset-url-config'
 
         begin
-          asset_uri = resolve(path, compat: false)
+          asset_uri = resolve(path)
         rescue FileNotFound
           # TODO: eh, we should be able to use a form of locate that returns
           # nil instead of raising an exception.
@@ -39,5 +39,10 @@ module Sprockets
     config = env.context_class.config
     [config.relative_url_root,
     (config.asset_host unless config.asset_host.respond_to?(:call))]
+  end
+
+  # fallback to the default pipeline when using Sprockets 3.x
+  unless config[:pipelines].include? :debug
+    register_pipeline :debug, config[:pipelines][:default]
   end
 end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -24,10 +24,7 @@ class HelperTest < ActionView::TestCase
     @view.assets_environment = @assets
     @view.assets_manifest    = @manifest
     @view.assets_prefix      = "/assets"
-    @view.assets_precompile  = %w(
-      foo.css foo.js bar.css bar.js
-      file1.css file1.js file2.css file2.js
-    )
+    @view.assets_precompile  = %w( manifest.js )
     @view.request = ActionDispatch::Request.new({
       "rack.url_scheme" => "https"
     })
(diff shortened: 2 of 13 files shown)


rafaelfranca:
Awesome. Great work @Andreis13!

rafaelfranca:
One thing that we can do on this PR is to set the travis matrix to run tests for the sprocket 4.x branch. What you will need is a new Gemfile like this one https://github.com/rails/sprockets-rails/blob/7bd1560360248ab11d8aba81e558bd817935c692/gemfiles/Gemfile.rails-5.0.x that points to sprockets branch and to add the Gemfile on this list https://github.com/rails/sprockets-rails/blob/7bd1560360248ab11d8aba81e558bd817935c692/.travis.yml#L11.

Andreis13:
I've told Travis to test on Sprockets 4 too. ))

lib/sprockets/rails/context.rb
```diff
@@ -17,7 +17,7 @@ def compute_asset_path(path, options = {})
         @dependencies << 'actioncontroller-asset-url-config'
 
         begin
-          asset_uri = resolve(path, compat: false)
+          asset_uri = resolve(path)
```
arthurnn: Lets make sure we can remove this `compat` option. Sprockets 3.x docs says that it will return a String if no option given. https://github.com/rails/sprockets/blob/3.x/lib/sprockets/legacy.rb#L35-L36

test/test_helper.rb
```diff
@@ -497,35 +540,65 @@ def setup
...
-    assert_servable_asset_url "/assets/bar.self-#{@bar_self_js_digest}.js?body=1"
-    assert_servable_asset_url "/assets/dependency.self-#{@dependency_self_js_digest}.js?body=1"
-    assert_servable_asset_url "/assets/file1.self-#{@file1_self_js_digest}.js?body=1"
-    assert_servable_asset_url "/assets/file2.self-#{@file2_self_js_digest}.js?body=1"
+    assert_servable_asset_url "/assets/foo.debug-#{@foo_debug_js_digest}.js"
+    assert_servable_asset_url "/assets/bar.debug-#{@bar_debug_js_digest}.js"
+    assert_servable_asset_url "/assets/dependency.debug-#{@dependency_debug_js_digest}.js"
+    assert_servable_asset_url "/assets/file1.debug-#{@file1_debug_js_digest}.js"
+    assert_servable_asset_url "/assets/file2.debug-#{@file2_debug_js_digest}.js"
+    else
```
arthurnn: indentation issue here.

lib/sprockets/rails/context.rb
Andreis13: The `compat: false` option can be safely removed because the `resolve` method that is present here is actually [Context#resolve](https://github.com/rails/sprockets/blob/3.x/lib/sprockets/context.rb#L88) and **not** [Environment#resolve](https://github.com/rails/sprockets/blob/3.x/lib/sprockets/resolve.rb#L23). It calls, however, the [Environment#resolve!](https://github.com/rails/sprockets/blob/3.x/lib/sprockets/resolve.rb#L52) which, as you can see in the highlighted line, enforces the `compat: false` option anyway, in Sprockets 3.

On the other hand, we cannot remove the `begin...rescue` block, due to the same `resolve!` method which raises the *FileNotFound* exception.

arthurnn:
Can you squash the commits in one before I merge it?
thanks

PR #261, opened 2015-06-27
Restore logger config (config.assets.logger)

edslocomb:
Not sure when this was removed, but I assume it was inadvertent since it's still in the docs.

At present, config.assets.logger is not being used by sprockets; nothing will be written to the logger you provide in your config.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -54,6 +54,10 @@ module Sprockets
       end
     end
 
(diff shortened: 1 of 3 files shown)


rafaelfranca:
You can config the logger using the `configure` blocks.

```
Rails.application.config.assets.configure do |env|
  env.logger = foo
end
```

We don't want to provide more options since it can be done with the existent one.

Thank you for the pull request.

PR #265, opened 2015-07-28
Fix performance regression in Rails helper

eileencodes:
This fixes the regression reported in #234 that @dhh reported. As noted in the commit message that code was a red herring. The problem was that the cache on `precompiled_assets` was being reset every single time by CachedEnvironment because it didn't pull `precompiled_assets` over.

This change moves that method to the Railtie class so that it remains in the env. 

If you want to test this out I made a brand new application to demonstrate the issue and test with (https://github.com/eileencodes/perf_regression_sprockets). 

This regression shows up on running tests that render assets - show that renders a template with application.js and application.css.

cc/ @jeremy 

---

Fix performance regression in Rails helper

Previously reported performance regression #234 was a red herring. The
`Engine` class and addition of adding `app/` asset path appears to be a
big problem, but the reason it's slow is because then it would run
through `precompiled_assets` where without that code it skips
`precompiled_assets` thus skipping the code that is actually slow.

Although `precompiled_assets` was memoized the cache wasn't actually
working and the method cache was being thrown away every single time.
This became very obvious when our test suite (Basecamp) went from 54
seconds to 320 seconds to run on sprockets-rails master.

The problem is that when the env is reset the `precompiled_assets` gets
reset instead of keeping the cache. I moved this method onto the config
to behave like the other methods, for example `assets_manifest`.

Benchmarks:
On a brand new application with a single show test that simply renders a
view with a few assets.

Before this change:
```
     Running test...    27.000  i/100ms
     Running test...    305.554  (± 8.2%) i/s -      1.539k
```

After this change:
```
     Running test...    55.000  i/100ms
     Running test...    558.135  (± 5.4%) i/s -      2.805k
```

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -21,7 +21,7 @@ module Sprockets
       include Sprockets::Rails::Utils
 
       VIEW_ACCESSORS = [:assets_environment, :assets_manifest,
-                        :assets_precompile,
+                        :assets_precompile, :precompiled_assets,
                         :assets_prefix, :digest_assets, :debug_assets]
 
       def self.included(klass)
@@ -226,11 +226,6 @@ module Sprockets
           end
 
           asset
-        end
-
-        # Internal: Generate a Set of all precompiled assets logical paths.
-        def precompiled_assets
-          @precompiled_assets ||= assets_manifest.find(assets_precompile || []).map(&:logical_path)
         end
     end
   end
--- a/lib/sprockets/rails/utils.rb
+++ b/lib/sprockets/rails/utils.rb
@@ -6,6 +6,11 @@ module Sprockets
       def using_sprockets4?
         Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new('4.0.0')
       end
+
+      # Internal: Generate a Set of all precompiled assets logical paths.
+      def build_precompiled_list(manifest, assets)
+        manifest.find(assets || []).map(&:logical_path)
+      end
     end
   end
 end
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -26,6 +26,9 @@ module Rails
 
     # Returns Sprockets::Manifest for app config.
     attr_accessor :assets_manifest
+
+    # Returns array of already precompiled assets
+    attr_accessor :precompiled_assets
   end
 
   class Engine < Railtie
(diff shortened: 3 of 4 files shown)


eileencodes:
Doh! forgot to mention this is present on:

sprockets-rails master but not sprockets-rails 2.x (this change actually makes it faster than 2.x was even though the perf regression wasn't as bad)

lib/sprockets/rails/helper.rb
```diff
@@ -53,6 +53,11 @@ def assets_environment
               nil
             end
           end
+
+          remove_method :precompiled_assets
+          def precompiled_assets
```
rafaelfranca: I could not figure out why we are defining this method here. Could you explain?

eileencodes: Yes sorry - I meant to do that.

So I'm not sure this is the "right way" to do this but in the test here: https://github.com/rails/sprockets-rails/blob/master/test/test_helper.rb#L730 does not run through the Railtie so `precompiled_assets` doesn't contain the url.css it's looking for.

I'm guessing it has to do with defining `@view.assets_precompile = ["url.css"]` in the test where url.css isn't available because it doesn't run through the Railtie. Maybe there's a better way to fix  that?

eileencodes: Ah, I haven't really worked with sprockets-rails before so i was trying not to change the tests in fixing this.

Thinking maybe I'm supposed to just call `precompile_assets` after the `@view.assets_precompile = ["url.css"]`?

rafaelfranca: I think we should just set the `@view.precompiled_assets` after the `@view.assets_precompile =` calll.

eileencodes: Ok fixed :+1:

rafaelfranca:
@Andreis13 has some considerations about this feature but the patch looks good. :shipit: but beware that it still has a bug when the logical path is not the same as the asset path.

eileencodes:
I introduced a bug or there is an existing bug?

rafaelfranca:
It is an existing bug with sprockets 4.

eileencodes:
Ah ok, cool. Just wanted to double check. I'll merge. Thanks for reviewing :smile_cat:

schneems:
This introduced a regression in the rails suite on master: https://travis-ci.org/rails/rails/jobs/73218179#L389 not sure why. I'm able to reproduce by using

```
gem 'sprockets-rails', github: 'rails/sprockets-rails', branch: 'master', :ref => 'ad4a43bd1bb19c86a8bf94a2ad5e477686161490'
```

In the rails test suite and running

```
bundle exec rake test:regular  TEST=test/application/assets_test.rb
```

Specifying the previous commit fixes the problem:

```
gem 'sprockets-rails', github: 'rails/sprockets-rails', branch: 'master', :ref => '6389649358846f37507206885e3ece145862a0be'
```

I'm looking into why this started initializing a UglifierCompressor instance, ping me on campfire if you have any leads.

PR #271, opened 2015-09-02
Add config.assets.cache_path

jordan-brough:
Make it easy to customize the cache path without having to re-specify
the whole FileStore.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -74,6 +74,7 @@ module Sprockets
     config.assets.compile     = true
     config.assets.digest      = true
     config.assets.cache_limit = 50.megabytes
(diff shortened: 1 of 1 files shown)


jordan-brough:
@arthurnn any thoughts on this? I rebased on master last week to fix some unrelated spec failures.

arthurnn:
Not sure if we wanna add new options.. Specially if we can do this already with another one... 
I am, usually, not a big fan of lots of options specially here..

@rafaelfranca @schneems what do you guys think?

rafaelfranca:
I'm with @arthurnn here. Specifying the whole cache store is the way to go. I don't see a reason to have a way to specify the cache store and also the cache path.

PR #273, opened 2015-09-18
Fix SRI example hash in README

mastahyeti:
The `integrity` attribute doesn't include a dash in the digest name.


arthurnn:
thanks @mastahyeti

PR #275, opened 2015-09-25
Speed up app boot by deferring asset precompile

jeremy:
* Lazy-load precompiled asset list so it doesn't block boot time
* Memoize it on the app so it doesn't get flushed along with AV instances
* References #265 and #234.

cc @eileencodes

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -21,7 +21,7 @@ module Sprockets
       include Sprockets::Rails::Utils
 
       VIEW_ACCESSORS = [:assets_environment, :assets_manifest,
-                        :assets_precompile, :precompiled_assets,
+                        :assets_precompile, :precompiled_asset_checker,
                         :assets_prefix, :digest_assets, :debug_assets]
 
       def self.included(klass)
@@ -82,7 +82,7 @@ module Sprockets
         if environment = assets_environment
           if asset = environment[path]
             unless options[:debug]
-              if !precompiled_assets.include?(asset.logical_path)
+              if !precompiled_asset_checker.call(asset.logical_path)
                 raise AssetNotPrecompiled.new(asset.logical_path)
               end
             end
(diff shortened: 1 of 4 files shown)


matthewd:
Does this move the load to after workers get forked, where that's relevant?

eileencodes:
I double checked that this didn't cause a test speed regression and they are about the same speed so that's good! :+1:

Benchmarks from Basecamp

Before #265 
```
     Running test...     1.000  i/100ms
     Running test...      0.639  (± 0.0%) i/s -      4.000  in   6.307078s
```

After #265 which caused a perf regression in boot time but sped up test time:
```
     Running test...     3.000  i/100ms
     Running test...     33.113  (±12.1%) i/s -    165.000 
```

This PR:
```
     Running test...     1.000  i/100ms
     Running test...     33.781  (± 8.9%) i/s -    167.000 
```

rafaelfranca:
This should also fix the case were the assets depends on routes.

jeremy:
@matthewd It's lazy-loaded so it will pull the list after forking, not ideal. However, the precompiled-asset check is only for dev/test: kicks in when you're using the environment, not a manifest, and when you request debug assets.

@rafaelfranca Reference?

eileencodes:
@jeremy the routes problem is the same as the problem my commit caused - moving assets to boot time. This meant that if they relied on routes that they wouldn't work

See comments here: https://github.com/rails/sprockets-rails/commit/753e48eeb2920e0663b46e8e6ddc7bf8a0f72c35#commitcomment-13057681

jeremy:
Realizing we have some dead assumptions in the asset helpers now that we always have an asset manifest. We'll always take the manifest path here:
* https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/rails/helper.rb#L75-L92
* https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/rails/helper.rb#L106-L118

We need to flip these for them to make any sense. It should be buggy right now, causing stuck asset digests and integrity hashes in dev *if* you've ever precompiled, leaving behind a stale manifest.

PR #276, opened 2015-09-27
Fix stuck assets in dev due to a stale precompiled manifest

jeremy:
If an asset manifest is available, we check it for digests first even
if it's stale and a Sprockets environment is available.

We want to check the asset environment first and use the manifest only
when we're precompiled with no Sprockets environment available.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -73,21 +73,13 @@ module Sprockets
       #
       # Returns String path or nil if no asset was found.
       def asset_digest_path(path, options = {})
-        if manifest = assets_manifest
-          if digest_path = manifest.assets[path]
-            return digest_path
+        if assets_environment
+          if asset = assets_environment[path]
+            raise_unless_precompiled_asset asset.logical_path unless options[:debug]
+            asset.digest_path
           end
-        end
-
-        if environment = assets_environment
-          if asset = environment[path]
-            unless options[:debug]
-              if !precompiled_asset_checker.call(asset.logical_path)
-                raise AssetNotPrecompiled.new(asset.logical_path)
-              end
-            end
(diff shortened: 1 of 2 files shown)


rafaelfranca:
I think the previous behavior is better. There are cases were you have the manifest and also have the environment present and for these cases I believe it makes more sense to use the manifest first instead of the environment (my application rely on this behavior for example). This also matches the behavior of sprockets-rails 2.x.

jeremy:
That matches the behavior of older Sprockets if you had enabled digests, so if you used a manifest alongside a live environment in production.

The old digest helper only used the manifest if asset digests were enabled, and digests were off by default, so dev apps never took these paths. When `assets.digest` flipped to `true` in [Rails 4.2.0](https://github.com/rails/rails/commit/f369bcf9a0dba0a945ca6fe53343c042f54c1fcf) and [sprockets-rails 3](https://github.com/rails/sprockets-rails/commit/6f787e56c6be2fb499c303eebbbcc436ceeed1d5), we started returning stale assets in dev where they weren't intended or tested.

Perhaps we should be explicit about which we want to resolve against instead.

rafaelfranca:
> Perhaps we should be explicit about which we want to resolve against instead.

Yes. We do have a manifest alongside with a live environment in production so if we flip this by default we will always use the live environment to resolve the assets.

jeremy:
Cool, I'll switch the order back and make this explicit.

Curious, how/why do you use the live env?

rafaelfranca:
We do have a helper that inline SVGs images in the HTML and those SVGs are inside the assets folder. In order to inline the SVG content it needs to resolve the asset path and read the content of the asset. I'm planning to change this to use only the manifest but it will require a lot of work.

jeremy:
Done @ #278

PR #278, opened 2015-09-29
Restore manifest/env asset lookup order and make it configurable

jeremy:
* Revert env-first lookup order from #276 to support using both manifest and a live environment in production.
* Default to manifest first, env second. Skip manifest if debug enabled or digests disabled. Skip env if it's unavailable.
* Override with `config.assets.resolve_with = %i[ manifest environment ]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,6 @@ config.assets.version = 'v2'
 
 Defaults to `/assets`. Changes the directory to compile assets to.
 
-**`config.assets.manifest`**
-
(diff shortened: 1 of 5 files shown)


rafaelfranca:
:clap:

The test failure is weird though

jeremy:
Fix test fail on 1.9 due to using `%i[ … ]`

rafaelfranca:
Now `**options`

jeremy:
Got it

PR #285, opened 2015-11-12
Update the AssetNotPrecompiled error message for Sprockets 4

jonatack:
to explain how to remedy the error using the manifest file to link assets/directories/trees.


jonatack:
Ah, need to include Sprockets::Rails::Utils. Will update.

PR #310, opened 2016-01-16
If debug is true, async options should be ignored.

nateberkopec:
See commit message and https://github.com/rails/rails/pull/23012

I'm not sure I really did the best job on the tests here. Please direct me to a more appropriate place to add these tests. We should probably have a test to make sure that when debug is off, async still works, but because of the test's inheritance hierarchy I couldn't figure out the right place to do this.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -125,6 +125,7 @@ module Sprockets
         integrity = compute_integrity?(options)
 
         if options["debug"] != false && request_debug_assets?
+          options.delete("async")
           sources.map { |source|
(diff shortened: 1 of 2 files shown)


vipulnsward:
@nateberkopec this seems to break compile task output, ref: https://github.com/vipulnsward/sprockets-rails/blob/a955c164d7e55bc133d118acba69db1ae5a4e357/test/test_railtie.rb#L324 
This spec fails. Can you take a look?

nateberkopec:
No longer needed b/c of new sourcemapping behaviors in Sprockets 4

PR #315, opened 2016-01-22
Change default cache location back to original

schneems:
In this commit https://github.com/rails/sprockets-rails/commit/80164cb0b3eae7dc8a6efab7553a7a8ab4e20c10 the `assets` folder looks to be accidentally removed from the cache file store. This causes problems if you're specifically moving that folder between builds.


rafaelfranca:
:shipit:

schneems:
Just shipped 3.0.1

PR #316, opened 2016-01-22
Fix the precompile checker with bundle files

rafaelfranca:
Sprockets allow you to create bundles using `bundle_name/index.js` as the file name. If you declare `bundle_name` in your manifest you should not get an `AssetNotPrecompiled` error.


schneems:
:ship:-it

PR #323, opened 2016-02-01
Update README for integrity

alexmuller:
As far as I can figure out from the code and tests, adding `integrity: true` won't actually add the integrity hash to the tag.

I've updated the README to be less misleading :)


rafaelfranca:
Thank you for the pull request. The documentation was correct. See https://github.com/rails/sprockets-rails/blob/master/test/test_helper.rb#L399

alexmuller:
Ah thanks for clarifying, I was lost in the test file. Sorry to waste your time!

rafaelfranca:
My pleasure! It is not a waste of time if I'm helping someone :smile:

PR #327, opened 2016-02-18
Update railtie to ignore no-extension files

baweaver:
Updated the railtie to ignore files with no extensions, such as LICENSE, that cause sprockets to error out on assets compilation.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -60,7 +60,7 @@ module Sprockets
 
     LOOSE_APP_ASSETS = lambda do |logical_path, filename|
         filename.start_with?(::Rails.root.join("app/assets").to_s) &&
-        !%w(.js .css).include?(File.extname(logical_path))
+        !['.js', '.css', ''].include?(File.extname(logical_path))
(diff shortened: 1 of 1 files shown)


baweaver:
https://github.com/sstephenson/sprockets/issues/347

rafaelfranca:
Why do you have a file like LICENCE in the app/assets folder?

rafaelfranca:
If it is because you are putting a vendor project inside app/assets you should move to `vendor/assets` and looking your last comment in the linked issue it is what is happening.

baweaver:
Bower-rails seems to have a nasty habit of doing that. That being said, it's a concern of both components as breaking on a file with no-extension feels like a bug that should be addressed regardless.

matthewd:
@rafaelfranca we're not the filename-police... if our code is going to barf on an extensionless file, it seems the least we can do is ignore it

rafaelfranca:
@matthewd yeah, agree. Thank you for double checking. BTW, this behavior is soft deprecated already and you should migrate to use the manifest file.

PR #329, opened 2016-02-23
Better Errors

schneems:
Raise a descriptive error if specific config file `app/assets/config/manifest.js` is not found when using Sprockets 4.

The `link ../javascripts/application.js` link does not work for me on codetriage.com but `link application.js` does.

The error when a source isn't found should only show how to add that specific source for maximum clarity.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -66,6 +66,20 @@ module Sprockets
   class Railtie < ::Rails::Railtie
     include Sprockets::Rails::Utils
 
+    class ManifestNeededError < StandardError
+      def initialize
+        msg = "Expected to find a manifest file in `app/assets/config/manifest.js`\n" +
+        "But did not, please create this file and use it to link any assets that need\n" +
+        "to be rendered by your app:\n\n" +
+        "Example:\n" +
+        "  //= link_tree ../images\n"  +
+        "  //= link_directory ../javascripts .js\n" +
+        "  //= link_directory ../stylesheets .css\n"  +
+        "and restart your server"
+        super msg
+      end
+    end
+
     LOOSE_APP_ASSETS = lambda do |logical_path, filename|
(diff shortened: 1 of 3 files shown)


lib/sprockets/railtie.rb
```diff
@@ -38,7 +38,10 @@ def asset_precompiled?(logical_path)
     # boot time, but ensure we cache the list so we don't recompute it for each
     # request or test case.
     def precompiled_assets
-      @precompiled_assets ||= assets_manifest.find(config.assets.precompile).map(&:logical_path).to_set
+      @precompiled_assets ||= begin
+        raise ManifestNeededError if Sprockets::Railtie.using_sprockets4? && !Rails.root.join("app/assets/config/manifest.js").exist?
```
rafaelfranca: Is not better to raise inside the Railtie, right before `precompile` being set?

schneems: `Rails.root` isn't defined when `config.assets.precompile`  is set.

rafaelfranca: Move that code to an initializer and it should work.

schneems: :+1: i'll try that

matthewd:
What happened to "it's backwards compatible" / "you can still use the precompile list"? :confused:

schneems:
Sprockets 3 is backwards compat release. Sprockets 4 is when the shins get removed.

schneems:
*shims*

rafaelfranca:
> What happened to "it's backwards compatible" / "you can still use the precompile list"?

It is still true, just don't put complex objects on it :smile:

schneems:
We should also add the "images" directory to the generated and recommended manifest.js

rafaelfranca:
:+1:

schneems:
Updated

PR #331, opened 2016-02-24
[Close #321] Clear Asset precompile cache on request

### LINKED ISSUE #321: Adding new images requires a server restart in development
There's actually two problems here.  First is that in development mode, one should be able to add images without consciously having to restart the server.  The second is that the error message that is produced if you either forget or are unaware of the requirement to restart the server isn't particularly helpful:

    Sprockets::Rails::Helper::AssetNotPrecompiled in Products#index

    Showing /home/rubys/git/awdwr/edition4/work-223/depot/app/views/products/index.html.erb where line #8 raised:

    Asset was not declared to be precompiled in production.
    Add `Rails.application.config.assets.precompile += %w( cs.jpg )` to `config/initializers/assets.rb` and restart your server

schneems:
When in debug mode we want to see if any new files have been added to our precompile list by nature of them being added to disk. For example if you add an image  to `images` and you have linked the directory:

```
//= link_directory ../images
```

This change also eliminates the need to restart the server in development when changes are made when using a manifest.js and Sprockets 4.

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -8,6 +8,7 @@ require 'sprockets/rails/context'
 require 'sprockets/rails/helper'
 require 'sprockets/rails/route_wrapper'
 require 'sprockets/rails/version'
+require 'sprockets/rails/flush_cache_middleware'
 require 'set'
 
 module Rails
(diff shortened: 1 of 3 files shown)


lib/sprockets/railtie.rb
```diff
@@ -40,9 +41,20 @@ def asset_precompiled?(logical_path)
...
     end
+
+    def clear_precompiled_assets_cache!
+      @precompiled_assets = nil
+    end
   end
 
   class Engine < Railtie
+    initializer :flush_precompiled_assets_cache do |app|
+      if app.config.assets.debug
```
matthewd: I think `cache_classes` is probably a more accurate approximation of "it's safe to cache things long term", despite it's overly narrow name.

lib/sprockets/railtie.rb
```diff
@@ -40,9 +41,20 @@ def asset_precompiled?(logical_path)
     def precompiled_assets
       @precompiled_assets ||= assets_manifest.find(config.assets.precompile).map(&:logical_path).to_set
     end
+
+    def clear_precompiled_assets_cache!
+      @precompiled_assets = nil
```
matthewd: Instead of actually flushing the cache on every request, can we just mark it "okay to flush", and then only *do* the flush if `asset_precompiled?` is about to return false?

schneems: We could actually do that and not need middleware. That's a good idea. I'll do that in a separate PR.

PR #332, opened 2016-02-24
[Close #321] Double check precompile missing asset

### LINKED ISSUE #321: Adding new images requires a server restart in development

schneems:
When in classes are not cached, we're not in a state where performance matters as much (like development mode) we want to see if any new files have been added to our precompile list by nature of them being added to disk. For example if you add an image  to `images` and you have linked the directory:

```
//= link_directory ../images
```

This change also eliminates the need to restart the server in development when changes are made when using a manifest.js and Sprockets 4.

cc/ @matthewd

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -18,8 +18,7 @@ module Sprockets
             "`//= link_directory ../javascripts .js`\n" +
             "`//= link_directory ../stylesheets .css`\n" +
             "`//= link_tree ../javascripts .js`\n" +
-            "`//= link_tree ../images`\n" +
-            "and restart your server"
+            "`//= link_tree ../images`\n"
           else
             "Asset was not declared to be precompiled in production.\n" +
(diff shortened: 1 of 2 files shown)


matthewd:
:ok_hand:  :heart:

rafaelfranca:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #341, opened 2016-04-15
Ensure precompiled asset check uses the proper asset path

liseki:
Within Rails a situation where ```asset.self.js``` is checked for precompilation may not be necessary or likely but I ran into such a situation with the integration with [Teaspoon](https://github.com/modeset/teaspoon).

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -327,7 +327,7 @@ module Sprockets
 
         def find_debug_asset(path)
           if asset = find_asset(path, pipeline: :debug)
-            raise_unless_precompiled_asset asset.logical_path.sub('.debug', '')
+            raise_unless_precompiled_asset asset.logical_path
             asset
           end
         end
@@ -342,6 +342,7 @@ module Sprockets
           end
 
           def raise_unless_precompiled_asset(path)
+            path = path.sub(/\.(debug|self)/, '')
             raise Helper::AssetNotPrecompiled.new(path) unless precompiled?(path)
           end
       end


liseki:
Ping @schneems

liseki:
@rafaelfranca could you take a look please?

schneems:
So this is a bit backwards, but could you open an issue that describes exactly what you're trying to fix? Maybe an example app that shows the problem? I've not encountered the failure case you're talking about before.

liseki:
Hi @schneems. Please let me know if my explanation can suffice. I am working to get [Teaspoon](https://github.com/modeset/teaspoon) working with Sprockets 4. Suppose we have a test file, first_test.js:
```
/*** first_test.js ***/
//= require asset_a
//= require support/asset_b

describe('...', function () {
  // Tests here ...
})
```
This test in Teaspoon can be run in two ways. In the first and general way Teaspoon loads the test file in the browser and runs the test, pretty straight forward and works as long as all the test file is declared in the precompile list. Teaspoon will look for the debug asset ```first_test.debug.js``` and Sprockets gives it ```first_test.debug-***.js```.

The second way of running tests involves testing coverage. To get meaningful results Teaspoon loads individual files so we can get individual metrics for each. In this case Teaspoon will look for ```asset_a.self.js```, ```support/asset_b.self.js``` and ```first_test.self.js``` and Sprockets should give it ```asset_a.self-***.js```, ```support/asset_b.self-***.js``` and ```first_test.self-***.js```.

This first approach works because [sprockets-rails](https://github.com/rails/sprockets-rails/blob/89ba006950caba495954fbb1626919bfe4ee7838/lib/sprockets/rails/helper.rb#L330) knows to strip out 'debug' from the filename when checking the precompile list. However the second approach, which I am working to fix with this patch, fails because [sprockets-rails](https://github.com/rails/sprockets-rails/blob/89ba006950caba495954fbb1626919bfe4ee7838/lib/sprockets/rails/helper.rb#L319) does not strip out 'self' from the filename. We have ```asset_a.js``` in the precompile but you'll agree that having to also add ```asset_a.self.js``` to he precompile list makes no sense. The 'debug' case is used by Rails and there are tests for it but Rails does not seem to need to load a 'self' asset. So this is a patch to help Teaspoon and other plugins that may need to load a 'self' asset.

liseki:
@schneems if this sounds reasonable to you perhaps I should add a test for it?

liseki:
Seems I may be able to get around this with the 'debug' and 'allow_non_precompiled' options to ```javascript_include_tag```. Closing this for now.

Issue #297, opened 2015-12-21
Registering assets for precompilation from within gem not working.


jejacks0n:
Hi, I'm the author of Teaspoon, a javascript test runner for rails, and a new issue showed up recently around some of the assets not being available in "production". Teaspoon doesn't use the production env, so that part is a bit confusing, and it does register its assets for precompilation.

It would appear at a glance that between sprockets v2* and v3 we're getting some breakage that's not fully clear. Hopefully someone here can shed some light as to why, so we can patch it up.

Thanks!

https://github.com/modeset/teaspoon/issues/443#issuecomment-166356580

rafaelfranca:
Cross posting my answer here:

@jejacks0n hi! I'm sprockets maintainer. What is happening is that teaspoon use [`javascript_tag` to include the users' defined test files](https://github.com/modeset/teaspoon/blob/cdcdc7d5d1608d6eaa9fc57e1b6e17552346fb1d/app/views/teaspoon/suite/_boot.html.erb#L1). Say you have the `test/javascript/unit/home_test.js` file. Teaspoon call `javascript_tag` with that as argument. That file is not registered for precompilation and sprockets-rails 3 now is strict about files that are being used as argument to `javascript_tag` and not being in the precompile list.

Does teaspoon maintain a list of all the test files? If so it is possible to register them?

I'm using teaspoon with sprockets-rails 3 and we are correctly doing to workaround this problem is:

```ruby
config.asset_manifest += [%r{^unit/}]
```

But I'd love to get ride of it since teaspoon already know which files are being called in `javascript_tag` and can register them.@jejacks0n hi! I'm sprockets maintainer. What is happening is that teaspoon use [`javascript_tag` to include the users' defined test files](https://github.com/modeset/teaspoon/blob/cdcdc7d5d1608d6eaa9fc57e1b6e17552346fb1d/app/views/teaspoon/suite/_boot.html.erb#L1). Say you have the `test/javascript/unit/home_test.js` file. Teaspoon call `javascript_tag` with that as argument. That file is not registered for precompilation and sprockets-rails 3 now is strict about files that are being used as argument to `javascript_tag` and not being in the precompile list.

Does teaspoon maintain a list of all the test files? If so it is possible to register them?

jejacks0n:
Thanks, I'll close this one and we can work out the details in the other issue. =)

jejacks0n:
@rafaelfranca I found what appears to be a fix, but I wanted to run it past you first.

[...]

PR #347, opened 2016-05-09
Disentangle 'debug' and 'allow_non_precompiled' asset path options

liseki:
This will allow us to close [issue 297](https://github.com/rails/sprockets-rails/issues/297) that sprung out of integrating Teaspoon with Rails. The crux of the issue is that attempting to load any asset not listed for pre-compilation results in an ```Sprockets::Rails::Helper::AssetNotPrecompiled``` error.

To recreate the issue attempt loading an asset not declared from pre-compilation like so:
```
<%= javascript_include_tag 'dummy', 'debug' => false, 'allow_non_precompiled' => true %>
```
with the following setup:
```
rails 4.2.6
sprockets 3.6.0
sprockets-rails 3.0.4
```

Internally sprockets-rails already has all the flags for loading an asset this way but the 'allow_non_precomiled' flag is ignored. This fixes that will falling back on the 'debug' flag when 'allow_non_precomiled' is not declared so that users can still rely on the old behaviour.

Please not that for the call to ```javascript_include_tag``` to actually work will additionally require ```ActionView::Helpers::AssetTagHelper#javascript_include_tag``` to pass along the 'allow_non_precompiled' (and 'debug') flags when generating the asset path. That will be dealt with as a second step. With this fix at least the following can now work:
```
<%= javascript_include_tag path_to_javascript('dummy', debug: false, allow_non_precompiled: true) %>
```

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -70,8 +70,10 @@ module Sprockets
 
       def compute_asset_path(path, options = {})
         debug = options[:debug]
+        allow_non_precompiled = options.key?(:allow_non_precompiled) ?
+          options[:allow_non_precompiled] : debug
 
-        if asset_path = resolve_asset_path(path, debug)
+        if asset_path = resolve_asset_path(path, allow_non_precompiled)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
           super


liseki:
There is a failing test on travis but I do not think that should happen because of this patch. I am going to re-attempt the tests.

liseki:
@rafaelfranca just checking your thoughts on this ...

rafaelfranca:
Thank you for  the pull request. In the next few days I'll be reviewing this whole feature so maybe we will not need to. I'll keep you updated.

liseki:
Thanks @rafaelfranca

rafaelfranca:
In the master branch we have a option to disable the precompile check now. Does that fix your issue?

liseki:
[Commit b4f9006c](https://github.com/rails/sprockets-rails/commit/b4f9006c7ff5ccd74b467c1b81e721c15dd5daf7) seems to take care of it. I'll write some tests over the next day or two just to confirm and then I can close this. Much appreciated.

liseki:
Generally what are your thoughts about the two approaches on how an engine gem can load an asset that is not declared for precompilation:

1. The global approach like what we now have on master by setting ```config.assets.check_precompiled_asset = false```. This makes perfect sense when the developer wants to disregard the asset exceptions in a particular environment, say development; any included engine gems now play along to that. However it feels a little heavy-handed in the opposite scenario when the developer has to set this option for his app because an included engine gem needs to load some assets.

2. The per-asset approach where the app or any included engine gem can load any asset in any particular state (e.g. 'source' or 'debug') as this patch was attempting to do. This has the benefit of having the asset exceptions when needed while providing the flexibility of loading any asset on file.

With your greater depth of knowledge on this I was just wondering about the background thinking on this.

liseki:
This has given us an option for fixing the issue we had in Teaspoon but in a rather compromised way. If you have a minute you can [see why here](https://github.com/modeset/teaspoon/issues/482#issuecomment-222375936). The global ```config.assets.check_precompiled_asset``` option is really handy and it would have been even nicer to couple that with a granular option that worked regardless of the current global config.

Thank you for your time on this!

PR #350, opened 2016-05-21
add option to make raising exception for missing assets optional

sbfaulkner:
@rafaelfranca this PR makes it possible to disable the exception for a missing asset (but still does by default)

tested in a real use-case and this does appear to eliminate the previously discussed first-request delay

let me know if this needs anything more

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ is not enabled, then we check the environment if compiling is enabled:
 ```
 If the resolver list is empty (e.g. if debug is true and compile is false), the standard rails public path resolution will be used.
 
+**`config.assets.raise_unless_precompiled_asset`**
+
+When enabled, an exception is raised for missing assets. This option is enabled by default.
+
 ## Complementary plugins
 
 The following plugins provide some extras for the Sprockets Asset Pipeline.
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ is not enabled, then we check the environment if compiling is enabled:
 ```
 If the resolver list is empty (e.g. if debug is true and compile is false), the standard rails public path resolution will be used.
 
-**`config.assets.raise_unless_precompiled_asset`**
+**`config.assets.check_precompiled_asset`**
 
 When enabled, an exception is raised for missing assets. This option is enabled by default.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -121,6 +121,10 @@ is not enabled, then we check the environment if compiling is enabled:
 ```
 If the resolver list is empty (e.g. if debug is true and compile is false), the standard rails public path resolution will be used.
 
+**`config.assets.raise_unless_precompiled_asset`**
```
rafaelfranca: Maybe `check_precompiled_asset`?

rafaelfranca:
Awesome! That is what I had in mind.

lib/sprockets/rails/helper.rb
```diff
@@ -342,6 +343,7 @@ def precompiled?(path)
           end
 
           def raise_unless_precompiled_asset(path)
+            return unless @raise_unless_precompiled_asset
             raise Helper::AssetNotPrecompiled.new(path) unless precompiled?(path)
```
kaspth: We could consolidate these conditionals:

```ruby
if @raise_unless_precompiled_asset && !precompiled?(path)
  raise Helper::AssetNotPrecompiled.new(path)
end
```

lib/sprockets/rails/helper.rb
```diff
@@ -299,6 +299,7 @@ def initialize(view)
           raise ArgumentError, 'config.assets.resolve_with includes :environment, but app.assets is nil' unless view.assets_environment
           @env = view.assets_environment
           @precompiled_asset_checker = view.precompiled_asset_checker
+          @raise_unless_precompiled_asset = view.raise_unless_precompiled_asset
```
kaspth: Alternatively, we could override `@precompiled_asset_checker` to scope the change to this place:

```ruby
@precompiled_asset_checker =
  if view.raise_unless_assets_precompiled
    view.precompiled_asset_checker
  else
    -> logical_path { true }
  end
```

sbfaulkner: that would make it so that the checker doesn't function at all, so I think we're better keeping the testing of the option out at the point that precompiled? is called (so that it doesn't cause side-effects down the road)

sbfaulkner:
@kaspth @rafaelfranca changes made based on (most of the) feedback... hows this now?

rafaelfranca:
Thanks!

sbfaulkner:
@rafaelfranca any chance we can get a release with this in it? (3.0.5 I think?)

rafaelfranca:
I'll do tomorrow.

PR #338, opened 2016-04-02
Bring quiet_assets to sprockets-rails

route:
As discussed here https://github.com/rails/rails/pull/24383 out of DHH comment here https://github.com/evrone/quiet_assets/issues/47#issuecomment-184186025
Again feel free to direct me and I'll fix everything that needs to be

PR #355, opened 2016-06-09
Silence asset requests with config.assets.quiet 

kmcphillips:
@rafaelfranca @josh 

Builds on and finishes what @route started in #338.

Basically a deprecation of the [`quiet_assets`](https://github.com/evrone/quiet_assets/) gem.

Adds the `config.assets.quiet` option. Defaulted to `false` in all cases. When true, it loads a middleware. That middleware looks at the asset prefix path, and if matches will wrap the given request in a `logger.silence { }` call.

I've added to the `README` but I don't know if/where I should document it further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ Add additional assets to compile on deploy. Defaults to `application.js`, `appli
 
 Add additional load paths to this Array. Rails includes `app/assets`, `lib/assets` and `vendor/assets` for you already. Plugins might want to add their custom paths to this.
 
+**`config.assets.quiet`**
+
+Suppresses logger output for asset requests. Uses the `config.assets.prefix` path to match and wraps the request in a `Rails.logger.silence{ }` which silences all but `ERROR` level. Defaults to `false`.
 
 **`config.assets.version`**
 
--- a/lib/sprockets/rails/quiet_assets.rb
+++ b/lib/sprockets/rails/quiet_assets.rb
@@ -1,0 +1,18 @@
+module Sprockets
+  module Rails
+    class QuietAssets
+      def initialize(app)
+        @app = app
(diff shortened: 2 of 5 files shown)


rafaelfranca:
Looks great. It seems the sprockets 4 tests are not passing. I guess it is because the tests you defined needs to point to a manifest file. Could you take a look?

kmcphillips:
Build fixed. 💚

sonalkr132:
Can we have version release please :blue_heart:

neohunter:
Using sprocket-rails 3.2.0 I'm still getting thinks like:

```
Started GET "/assets/img/icon/status2x.png" for ::1 at 2017-02-24 05:44:18 -0500
Started GET "/assets/plugins/jquery-datatable/images/sort_asc.png" for ::1 at 2017-02-24 05:44:19 -0500
```

This should be removed also? or this is from the webserver (thin)?

BattleBrisket:
According to my testing, the integration of the `quiet_assets` gem only applies to _asset_ logs, and not the _request_ logs for said assets. 

Can anyone provide some direction as to how I might eliminate the remaining logs lines above in Rails 5? I understand that the team [has](https://github.com/rails/rails/issues/4569#issuecomment-3594500) [repeatedly](https://github.com/rails/rails/pull/3795#issuecomment-3549669) [stated](https://github.com/rails/rails/issues/14608#issuecomment-39621797) that they will not disable these logs, I am just looking for a way to silence them in my own environments, and older solutions do not seem to work in the 5.x line.

Pysis868:
For these other messages, I found through the middleware stack that they are produced by WEBrick as access log messages.  I was finally able to quiet them using this Rails Server init config `boot.rb` method: https://github.com/rails/rails/issues/28968#issuecomment-327099731

PR #358, opened 2016-06-19
Cleanup documentation for quiet assets [ci skip]

prathamesh-sonpatki:


README.md
```diff
@@ -61,7 +61,7 @@ Add additional load paths to this Array. Rails includes `app/assets`, `lib/asset
 
 **`config.assets.quiet`**
 
-Suppresses logger output for asset requests. Uses the `config.assets.prefix` path to match and wraps the request in a `Rails.logger.silence{ }` which silences all but `ERROR` level. Defaults to `false`.
+Suppresses logger output for asset requests. Uses the `config.assets.prefix` path to match and wraps the request in a `Rails.logger.silence { }` which silences all asset requests but `ERROR` level. Defaults to `false`.
```
rafaelfranca: The previous documentation was correct. It says that it silences all log messages but the ones with the ERROR level. It is not talking about the assets at all. We should either change it to talk about the assets request, or make explicit that it is talking about log messages.

prathamesh-sonpatki: @rafaelfranca Thanks. I didn't realize  that `which` in `in a `Rails.logger.silence{ }` which`  is about `Rails.logger.silence`.  I removed `Rails.logger.silence { }` completely.

PR #360, opened 2016-06-20
Add a note that SRI only works over HTTPS

connorshea:
Add a note that the SRI attribute is only applied when the Rails app is served over HTTPS. This is something that tripped me up initially, as browsers treat localhost as a secure origin.

Source: https://github.com/rails/sprockets-rails/commit/fdb0df6b7bf890df3a632e754ec7d4e51688662f

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,8 @@ javascript_include_tag :application, integrity: true
 # => "<script src="/assets/application.js" integrity="sha256-TvVUHzSfftWg1rcfL6TIJ0XKEGrgLyEq6lEpcmrG9qs="></script>"
 ```
 
+Note that sprockets-rails only adds integrity hashes to assets when served over an HTTPS connection.
+
 
 ## Contributing to Sprockets Rails
 


README.md
```diff
@@ -158,6 +158,8 @@ javascript_include_tag :application, integrity: true
 # => "<script src="/assets/application.js" integrity="sha256-TvVUHzSfftWg1rcfL6TIJ0XKEGrgLyEq6lEpcmrG9qs="></script>"
 ```
 
+Note that sprockets-rails only adds integrity hashes to assets when served over an HTTPS connection.
```
vipulnsward: "assets, when"

connorshea: I don't think that'd be grammatically correct as it'd break up the statement into two clauses which don't make sense separately, "Note that sprockets-rails only adds integrity hashes to assets" and "when served over an HTTPS connection."

vipulnsward: True, it needs continuity here.

PR #362, opened 2016-06-23
Remove requirement of HTTPS for using SRI

connorshea:
The spec has been updated since this code was originally written and this no longer requires a Secure Context: https://www.w3.org/TR/SRI/#is-response-eligible-for-integrity-validation

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -178,22 +178,10 @@ module Sprockets
         # doesn't bleed into the tag attributes, but also check its value if
         # it's boolean-ish.
         def compute_integrity?(options)
-          if secure_subresource_integrity_context?
-            case options['integrity']
-            when nil, false, true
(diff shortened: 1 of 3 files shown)


lib/sprockets/rails/helper.rb
```diff
@@ -189,10 +189,11 @@ def compute_integrity?(options)
           end
         end
 
-        # Only serve integrity metadata for HTTPS requests:
-        #   http://www.w3.org/TR/SRI/#non-secure-contexts-remain-non-secure
+        # The SRI spec does not require HTTPS, although you're
+        # not benefiting from the extra security without HTTPS.
+        #   https://www.w3.org/TR/SRI/#is-response-eligible-for-integrity-validation
         def secure_subresource_integrity_context?
```
rafaelfranca: We can remove the method entirely because it is not checking anything.

connorshea:
@rafaelfranca good point, removed.

PR #375, opened 2016-08-19
Deprecate asset fallback

schneems:
When an asset isn't found the behavior is to pass the string through. For example a valid asset will return a url from pipeline

```
asset_path("application.js")
# => assets/application-123098udasvi0mnafd.js
```

While if you make a typo, you won't get an error or anything, it just falls through:

```
asset_path("app1icati0n.js")
# => "app1icati0n.js"
```

Hopefully I don't have to elaborate on why this is bad.

This PR is a child PR to one in Rails that will introduce a `public_asset_path` API.

There are valid reasons for not using the asset pipeline, if you have a purely static asset in the public folder or if you want to link to a static URL in your assets somewhere it makes sense to declare that intention. Eventually we will replace the behavior of the deprecation with an exception so people don't lose hours of their to typos.

We only emit the deprecation when a `public_` api is available in Rails. This means that sprockets-rails can still be used in a backwards compatible manner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,11 @@ Each asset task will invoke `assets:environment` first. By default this loads th
 
 Also see [Sprockets::Rails::Task](https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/rails/task.rb) and [Rake::SprocketsTask](https://github.com/rails/sprockets/blob/master/lib/rake/sprocketstask.rb).
 
-
 ### Initializer options
+
+**`config.assets.unknown_asset_fallback`**
+
+When set to a truthy value, the a result will be returned even if the requested asset is not found in the asset pipeline. When set to `false` it will raise an error.
 
 **`config.assets.precompile`**
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Also see [Sprockets::Rails::Task](https://github.com/rails/sprockets-rails/blob/
 
 **`config.assets.unknown_asset_fallback`**
 
-When set to a truthy value, the a result will be returned even if the requested asset is not found in the asset pipeline. When set to `false` it will raise an error.
+When set to a truthy value, a result will be returned even if the requested asset is not found in the asset pipeline. When set to a falsey value it will raise an error when no asset is found in the pipeline. Defaults to `true`.
 
 **`config.assets.precompile`**
 
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -7,6 +7,7 @@ module Sprockets
   module Rails
     module Helper
       class AssetNotFound < StandardError; end
+
       class AssetNotPrecompiled < StandardError
         include Sprockets::Rails::Utils
         def initialize(source)
@@ -82,8 +83,10 @@ module Sprockets
           raise AssetNotFound, message unless unknown_asset_fallback
 
           if respond_to?(:public_compute_asset_path)
-            message << "The public fallback behavior is being deprecaed and will be removed.\n"
-            message << "pass in `skip_pipeline: true` instead.\n"
+            message << "Falling back to an asset that may be in the public folder.\n"
+            message << "This behavior is deprecated and will be removed.\n"
+            message << "To bypass the asset pipeline and preserve this behavior,\n"
+            message << "use the `skip_pipeline: true` option.\n"
 
             call_stack = respond_to?(:caller_locations) ? caller_locations : caller
             ActiveSupport::Deprecation.warn(message, call_stack)
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -68,15 +68,35 @@ module Sprockets
         end
       end
 
+      # Writes over the built in ActionView::Helpers::AssetUrlHelper#compute_asset_path
+      # to use the asset pipeline.
       def compute_asset_path(path, options = {})
         debug = options[:debug]
 
         if asset_path = resolve_asset_path(path, debug)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
-          super
+          result = super
+          if respond_to?(:public_asset_path)
+            throw(:asset_not_found, result)
+          else
+            result
+          end
         end
       end
+
+      # Writes over the built in ActionView::Helpers::AssetUrlHelper#asset_path
+      # to use the asset pipeline.
+      def asset_path(*args)
+        catch_asset_not_found = catch(:asset_not_found) do
+          return super(*args)
+        end
+
+        result = catch_asset_not_found
+        deprecate_invalid_asset_lookup(result, caller)
+        result
+      end
+      alias_method :path_to_asset, :asset_path # aliased to avoid conflicts with an asset_path named route
 
       # Resolve the asset path against the Sprockets manifest or environment.
       # Returns nil if it's an asset we don't know about.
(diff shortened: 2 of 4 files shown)


schneems:
The Rails PR is https://github.com/rails/rails/pull/26226

lib/sprockets/rails/helper.rb
```diff
@@ -6,6 +6,7 @@
 module Sprockets
   module Rails
     module Helper
+      class AssetNotFound < StandardError; end
```
kaspth: Shouldn't we add a newline after this class definition?

lib/sprockets/rails/helper.rb
```diff
@@ -68,12 +70,24 @@ def assets_environment
...
 
         if asset_path = resolve_asset_path(path, debug)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
+          message =  "The asset #{ path.inspect } you are looking for is not present in the asset pipeline."
+          raise AssetNotFound, message unless unknown_asset_fallback
+
+          if respond_to?(:public_compute_asset_path)
+            message << "The public fallback behavior is being deprecaed and will be removed.\n"
+            message << "pass in `skip_pipeline: true` instead.\n"
```
kaspth: `Pass in...`?

lib/sprockets/rails/helper.rb
```diff
@@ -68,12 +70,24 @@ def assets_environment
...
         debug = options[:debug]
 
         if asset_path = resolve_asset_path(path, debug)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
+          message =  "The asset #{ path.inspect } you are looking for is not present in the asset pipeline."
+          raise AssetNotFound, message unless unknown_asset_fallback
+
+          if respond_to?(:public_compute_asset_path)
+            message << "The public fallback behavior is being deprecaed and will be removed.\n"
```
kaspth: If users see this message it's because we *have* deprecated this type of lookup. So I'd go with `is deprecated` instead (there's also a typo in deprecated).

lib/sprockets/rails/helper.rb
```diff
@@ -244,6 +258,19 @@ def legacy_debug_path(path, debug)
...
+
+      private
+        # Emits a deprecation warning when asset pipeline
+        # is used with an asset that is not part of the pipeline.
+        #
+        # Attempts to determine proper method name based on caller.
+        def deprecate_invalid_asset_lookup(name, call_stack)
+          message =  "The asset #{ name.inspect } you are looking for is not present in the asset pipeline.\n"
+          message << "The public fallback behavior is being deprecated and will be removed.\n"
+          message << "pass in `skip_pipeline: true` instead.\n"
```
kaspth: `Pass in...`?

README.md
```diff
@@ -48,9 +48,12 @@ Each asset task will invoke `assets:environment` first. By default this loads th
 
 Also see [Sprockets::Rails::Task](https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/rails/task.rb) and [Rake::SprocketsTask](https://github.com/rails/sprockets/blob/master/lib/rake/sprocketstask.rb).
 
-
 ### Initializer options
 
+**`config.assets.unknown_asset_fallback`**
+
+When set to a truthy value, the a result will be returned even if the requested asset is not found in the asset pipeline. When set to `false` it will raise an error.
```
kaspth: You mention truthy and yet also use an explicit `false` — shouldn't a falsy value work?

`the a result` -> `the result`.

lib/sprockets/rails/helper.rb
schneems: How is this?

```
            message << "Falling back to an asset that may be in the public folder.\n"
            message << "This behavior is deprecated and will be removed.\n"
            message << "To bypass the asset pipeline and preserve this behavior,\n"
            message << "use the `skip_pipeline: true` option.\n"
```

lib/sprockets/rails/helper.rb
schneems: done

lib/sprockets/rails/helper.rb
schneems: Not sure why this comment shows twice.

schneems: Ahh, i need to remove this method, we're not using it anymore. Added this in when we tried to get fancy computing exactly where in the call stack you were when trying to look up the asset.

lib/sprockets/rails/helper.rb
matthewd: I'd consider something more terse, e.g.:

```
"Asset fallback is deprecated. To bypass the asset pipeline, use `skip_pipeline: true`"
```

Because of where/how they show up (i.e., often many times in a row), overly-helpful deprecation messages can just be overwhelming. IMO to-the-point + Google-able are the most valuable attributes here.

lib/sprockets/rails/helper.rb
```diff
@@ -68,12 +70,24 @@ def assets_environment
...
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
+          message =  "The asset #{ path.inspect } you are looking for is not present in the asset pipeline."
+          raise AssetNotFound, message unless unknown_asset_fallback
+
+          if respond_to?(:public_compute_asset_path)
+            message << "The public fallback behavior is being deprecaed and will be removed.\n"
+            message << "pass in `skip_pipeline: true` instead.\n"
+
+            call_stack = respond_to?(:caller_locations) ? caller_locations : caller
```
matthewd: If we're on a Rails new enough to have `public_compute_asset_path`, that implies we're on a Ruby new enough to have `caller_locations`, doesn't it?

lib/sprockets/rails/helper.rb
```diff
@@ -68,12 +70,24 @@ def assets_environment
...
 
+      # Writes over the built in ActionView::Helpers::AssetUrlHelper#compute_asset_path
+      # to use the asset pipeline.
       def compute_asset_path(path, options = {})
         debug = options[:debug]
 
         if asset_path = resolve_asset_path(path, debug)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
+          message =  "The asset #{ path.inspect } you are looking for is not present in the asset pipeline."
```
matthewd: Don't use "you" in an exception message (or a deprecation).

I think this should have separate phrasings for the exception vs the deprecation, actually, even though the intent is similar. The deprecation wants a softer "we didn't find it in the place [but we're doing this other thing]" ('not present in'); for the exception, that feels too gentle... a more definitive "not found" / "unknown asset" seems called for.

README.md
matthewd: Does Sprockets have precedent for using 'truthy'? Rails uses 'true' (as opposed to '`true`') for that.

kaspth:
Does @matthewd's comment about `caller_locations` stand? Otherwise :shipit: 👏

lib/sprockets/rails/helper.rb
schneems: The tests don't actually use Rails 5.1, we define the method, and the tests get run on all Ruby versions. The alternative to this would be detecting ruby versions and skipping the tests, which is less clean. I also want to be able to make a gem so that older versions of Rails can use this flag to get the error/deprecation behavior.

rlue:
I know it's been a long time since this PR was merged. Is there any chance we could get an update to [the section of the Rails Guides that addresses asset fallback](http://edgeguides.rubyonrails.org/asset_pipeline.html#raise-an-error-when-an-asset-is-not-found)?

I'd be glad to submit a PR with the approval of the team (but otherwise, I'm reluctant to spend any time on it).

PR #383, opened 2016-11-09
Be more specific with required file

EiNSTeiN-:
This fixes a problem with being unable to load `ActionView::Helpers::AssetUrlHelper` from L7 in some situations

@rafaelfranca

### DIFF
--- a/lib/sprockets/rails/context.rb
+++ b/lib/sprockets/rails/context.rb
@@ -1,4 +1,4 @@
-require 'action_view'
+require 'action_view/helpers'
 require 'sprockets'
 
 module Sprockets


arthurnn:
not sure why CI is failing, any clue?

EiNSTeiN-:
It looks unrelated, unfortunately I can't re-run the ci build (or if I can, I can't find the button to do it)

rafaelfranca:
I was debugging why this happening. When we require the `action_view` top file it setup an autoload hook for `ActionView::Helpers` pointing to `action_view/helpers`. In @EiNSTeiN- situation seems like a gem is define `ActionView::Helpers` before sprockets-rails is loaded so `action_view/helpers` will never be loaded since we only load it when the constant is missing.

Not loading `action_view/helpers` will not setup the autoload hooks for the helpers inside that file so we will see the LoadError exception.

PR #388, opened 2017-01-29
append_assets_path should be called before finisher_hook

thebravoman:
Migrating to rails 5 context.
A project includes the refinerycms engine (example project prepared below).
When the project includes the refinerycms than the app assets are not added to Rails.application.assets.paths. From there non of the assets in the app/assets folder could be found.

There is an initializer called append_assets_path that gets called in rails4, in rails 5 without the engine and in rails 5 with the engine. The initializer is there.

But when the engine is present the append_assets_path is called after the finisher_hook initializer. This happens because the initializers are sorted in this way:

```
 def run_initializers(group=:default, *args)
      return if instance_variable_defined?(:@ran)
      initializers.tsort_each do |initializer|
        initializer.run(*args) if initializer.belongs_to?(group)
      end
      @ran = true
    end
```
Here the end result after initializers.tsort_each is
```
...
finisher_hook
...
append_assets_path
```

But i think the correct order should be

```
...
append_assets_path
...
finisher_hook
```
It is important to use the finisher_hook because it moves all the paths from config.assets.paths to env

Look at sprokects-rails-3.2.0/lib/sprockets/railtie.rb
```
config.assets.configure do |env|
  config.assets.paths.each { |path| env.append_path(path) }
end
```

Move information:

Initially  working on issue
https://github.com/refinery/refinerycms/issues/3178
From there created issue
https://github.com/rails/sprockets/issues/458

I have prepared a demo project
https://github.com/thebravoman/sprockets_not_loading_assets

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -54,7 +54,7 @@ module Rails
   class Engine < Railtie
     # Skip defining append_assets_path on Rails <= 4.2
     unless initializers.find { |init| init.name == :append_assets_path }
-      initializer :append_assets_path, :group => :all do |app|
+      initializer :append_assets_path, before: :finisher_hook, :group => :all do |app|
         app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
         app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)


rafaelfranca:
Is the order of the initializers correct without the refinery/refinerycms engine?

rafaelfranca:
This seems to be an issue in the `refinery/refinerycms` engine. When it did this https://github.com/refinery/refinerycms/blob/2c3b7c31314cc3fbcd425d7c461e14002bfda7fc/images/lib/refinery/images/engine.rb#L11, it moved the finish hooker to before it should be placed. They should fix their engines to make sure it doesn't change the Rails load order.

thebravoman:
But an engine being able to so drastically change Rails load order is very error prone. It is hard to debug, to detect, to maintain. Also finished_hook and append_assets_path are both coming from rails not from refinery. Is there a relation between them that should be strict or should we count on the tsort algorithm always sorting them correctly even if there are not relations between the two jobs?

bricesanchez:
Thanks for pointing us the problem, i will try to see if your solution fix the bug. :)

bsodmike:
Seems this wasn't merged? For those unfamiliar with the issue here, we've done a lot of spelunking on the Refinery side of things thanks to @parndt @bricesanchez and @thebravoman (and others) https://github.com/refinery/refinerycms/issues/3178

bsodmike:
Please see my attempt to check this here and the associated commit info https://github.com/refinery/refinerycms/issues/3178#issuecomment-292671360 CC @parndt @bricesanchez  @thebravoman

bricesanchez:
> This seems to be an issue in the refinery/refinerycms engine. When it did this https://github.com/refinery/refinerycms/blob/2c3b7c31314cc3fbcd425d7c461e14002bfda7fc/images/lib/refinery/images/engine.rb#L11, it moved the finish hooker to before it should be placed. They should fix their engines to make sure it doesn't change the Rails load order.

Hi @rafaelfranca,

Could you show me an example of what i should change on refinery, i not able to fix the problem and i'm not sure to do the right thing?

rafaelfranca:
I'm not sure how to fix that use case because I don't know the requirement but I'd try removing the `before` option in that initializer

bricesanchez:
I've found that the finisher `:set_routes_reloader_hook` is guilty on Rails 5.x

If i remove all `:set_routes_reloader_hook` code blocks, assets are back.
See : https://github.com/refinery/refinerycms/search?utf8=%E2%9C%93&q=set_routes_reloader_hook&type=

I don't understand what happens between Rails 4 and 5.

Do you have any idea @rafaelfranca? @parndt?

PR #394, opened 2017-03-13
Fix tests

schneems:
Tests were broken when the default paths that rails loads started to include an asset in action_view. Instead of testing for such brittle conditions, we can look for the elements we want instead.

### DIFF
--- a/test/test_railtie.rb
+++ b/test/test_railtie.rb
@@ -80,7 +80,8 @@ class TestRailtie < TestBoot
     assert_equal ROOT, env.root
     assert_equal "", env.version
     assert env.cache
-    assert_equal ["#{ROOT}/app/assets/config"], env.paths
+    assert_includes(env.paths, "#{ROOT}/app/assets/config")
+
     assert_nil env.js_compressor
     assert_nil env.css_compressor
   end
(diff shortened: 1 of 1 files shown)


test/test_railtie.rb
```diff
@@ -80,7 +80,8 @@ def test_defaults_to_compile_assets_with_env_and_manifest_available
     assert_equal ROOT, env.root
     assert_equal "", env.version
     assert env.cache
-    assert_equal ["#{ROOT}/app/assets/config"], env.paths
+    assert_array_includes_item(env.paths, "#{ROOT}/app/assets/config")
```
rafaelfranca: Does not minitest has `assert_includes`?

schneems:
Now i'm getting new errors

```
Gem::InstallError: nokogiri requires Ruby version >= 2.1.0.
An error occurred while installing nokogiri (1.7.0.1), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.7.0.1'` succeeds before bundling.
The command "eval bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle} " failed. Retrying, 3 of 3.
```

junaruga:
@schneems about your errors, you can fix with my PR's code https://github.com/rails/sprockets-rails/pull/399 . The code has been reviewed by @aried3r 
Though I am still seeing another issue on my PR..

So, feel free to use the code of my PR.

PR #395, opened 2017-03-15
Do not suggest `link_directory` in manifest.js

schneems:
The `link_directory` directive will compile an asset for EVERY file in that directory. To preserve the behavior of Rails 3 we should encourage only generating a minimal subset of assets such as `application.js` and application.css`.


schneems:
@rafaelfranca there was more discussion over at https://github.com/rails/rails/pull/28430

PR #404, opened 2017-05-29
Specify master branch for Travis build status image.

junaruga:
I changed build status image URL.
Maybe we want to check build status for only master branch. We do not want to see build status for PR.

I referred `rails/rails` build status.
https://github.com/rails/rails/blob/master/README.md

You can see modified README here.
https://github.com/junaruga/sprockets-rails/blob/feature/readme-badge-master/README.md


schneems:
thanks!

PR #414, opened 2017-11-17
Schneems/more error stuff

schneems:


ghiculescu:
This sort of had a regression which I've fixed in https://github.com/rails/sprockets-rails/pull/472

PR #417, opened 2017-12-26
Configurable file extensions that are not copied to public

renra:
https://github.com/rails/rails/issues/5757


renra:
I think there's no need to keep this open anymore.

Issue #343, opened 2016-04-26
What was the resolution of #49 regarding non-digest assets?


joshjordan:
@rafaelfranca you made a comment here: https://github.com/rails/sprockets-rails/issues/49

...but that thread still has a lot of people flowing to it, and they don't know where to go. Your comment doesn't explain what the resolution or plan is, so over two years later, we are still working around this gems like https://github.com/alexspeller/non-stupid-digest-assets

Can you please give us an update, and make sure that thread sees it? Many places on the web about the issue link to that thread.

rafaelfranca:
There is no update. It is something in our radar but not a priority.

joshjordan:
That isn't consistent with your previous post:

```
We already have enough information to think about this issue and in fact we are going to handle it in some way (we already did some improvements on master branch to alleviate the pain)
```

What are those improvements? How do we take advantage of them?

rafaelfranca:
The improvements were raising an exception when a non-digest asset is referenced. You don't need to do anything to take advantage of it, it is enabled by default.

joshjordan:
I see. That is not the behavior we experience on 2.3.3. Even if it were, I do not see how that addresses the problem.

What can a contributor do to make progress on this? You mentioned that it is on the radar but not a priority, which is a good case for a contributor to take on.

rafaelfranca:
https://github.com/rails/sprockets-rails/pull/239

joshjordan:
That is an open pull request. What can a contributor do to make progress on this?

joshjordan:
@celsodantas do you understand what the next steps are here?

rafaelfranca:
It depends which problem are we trying to solve. The problem we are trying to solve is: there is some assets that we don't want to generate digest. But there are people that want: we want to generate a version of each asset in our application that don't have a digest.

We want to solve the first problem, but we don't want to solve the second problem. #49 talks about both. Which problem are you expecting to be solved?

joshjordan:
I'm really asking what _you_ were talking about when you said:

> in fact we are going to handle it in some way

1. Please, tell us what that way is so that we can implement it.
2. Can you explain what your intention was by linking PR #239?

cc @alexspeller

rafaelfranca:
The way is #239.

joshjordan:
Then I guess I'll just ask the same question I asked yesterday:

That is an open pull request. What can a contributor do to make progress on this? Why isn't it merged?

rafaelfranca:
[...]

PR #428, opened 2018-06-07
Document options for generating non-digest assets

mvz:
This documents the options suggested in https://github.com/rails/sprockets-rails/issues/343#issuecomment-215241277 and https://github.com/rails/sprockets-rails/issues/343#issuecomment-215234763 so people can hopefully find them more quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ The following plugins provide some extras for the Sprockets Asset Pipeline.
 You have several options:
 
 * Use the [non-digest-assets gem](https://github.com/mvz/non-digest-assets).
+* Use the [sprockets-redirect gem](https://github.com/sikachu/sprockets-redirect).
+* Use the [smart_assets gem](https://github.com/zarqman/smart_assets).
 * Create [a rake task](https://github.com/rails/sprockets-rails/issues/49#issuecomment-20535134) to pre-generate a non-digest version in `public/`.
 
 ## Experimental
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -150,6 +150,13 @@ The following plugins provide some extras for the Sprockets Asset Pipeline.
 * `config.assets.manifest` (if used) must now include the manifest filename, e.g. `Rails.root.join('config/manifest.json')`. It cannot be a directory.
 * Two cleanup tasks: `rake assets:clean` is now a safe cleanup that only removes older assets that are no longer used, while `rake assets:clobber` nukes the entire `public/assets` directory. The clean task allows for rolling deploys that may still be linking to an old asset while the new assets are being built.
 
+### But what if I want sprockets to generate non-digest assets?
+
+You have several options:
+
+* Use the [non-stupid-digest-assets gem](https://github.com/alexspeller/non-stupid-digest-assets).
```
rafaelfranca: Sorry, but I really don't want to recommend a gem with that name. Can you remove it?

mvz: I suppose I could fork it and give it a better name. It doesn't look that well-maintained either.

Do you still think a gem with that functionality is a reasonable solution?

mvz: I'm currently researching some other options and will update this PR accordingly.

mvz: Updated.

rafaelfranca: yes, I do think that gem is a reasonable solution

rafaelfranca:
Thanks!

mvz:
Thanks for merging this @rafaelfranca.

PR #434, opened 2018-09-25
Fix most travis failures, test against ruby 2.5.1

lostapathy:


lostapathy:
This currently fixes most, but not all, of the travis build failures.  I am unable to duplicate the builds that are currently failing - if someone has pointers on that I will be glad to work on it.

Worth noting that the "rails-5.1" builds were actually attempting to build edge rails, which has been fixed and targets added for rails-5.2 as well.

f58a451d6c4d2a8f5ec42682b0dc0ebbded8cd1a appears to be more than just a warning fix.  Prior to that, I would see at least one build in each run fail with odd messages like [this](https://travis-ci.org/rails/sprockets-rails/jobs/433109325). It seems related to github throttling requests to git:// urls, but I couldn't find anything definitive on it.

sprockets-rails.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency "sprockets", ">= 3.0.0"
   s.add_dependency "actionpack", ">= 4.0"
   s.add_dependency "activesupport", ">= 4.0"
+  s.add_development_dependency "minitest", "< 5.11.0"
```
rafaelfranca: Do we need to lock the minitest version for all build or only for some?

lostapathy: Not quite all, but most of them - see [before](https://travis-ci.org/rails/sprockets-rails/builds/433058147) and [after](https://travis-ci.org/rails/sprockets-rails/builds/433095502) on that.

I'm open to changing that if it gets this merged.  I'd rather merge it like this, sort out the remaining failing builds, then rework the tests so they work with newer minitest and then we could back this out.  At this point I feel like there are too many failing builds to start changing tests to resolve this.

lostapathy: Looking at this some more, I have figured out the other failures and will add a commit for those fairly shortly.  It doesn't look like the minitest issue is fixable (root cause in another gem) so I will move that from the gemspec into individual gemfiles after all.

rafaelfranca: I mea, I think we should specific the minitest restriction only in the Gemfiles that fail with the latest. Or it is all Gemfiles that fail?

lostapathy:
Thanks for the review @rafaelfranca   I went ahead and changed the minitest restriction to just the affected gemfiles - you're right, that makes more sense, especially since we won't be able to remove that restriction later.

Some test failures seem to be execution order dependent with the sprockets-4.0 betas.  I added a workaround in 90dcb70 for the most consistent ones, but I think other "random" test failures are possible on future travis builds.  I will try to isolate those and submit bug/PR with sprockets project in the next day or two.

rafaelfranca:
Thank you for this PR. Really good work.

PR #445, opened 2019-10-09
Use right sprockets version for backward compatibility

savin-alexey:


rafaelfranca:
Thank you for the full request. This gem supports both sprockets 3 and sprockets 4, so its gemspec is correct. If you application don't work with sprockets 4 you need to either fix your application or lock sprockets in your Gemfile.

Issue #369, opened 2016-07-28
sprockets 4 manifest.js location


dreyks:
Is there a plan to make  the location of `manifest.js` customizable?

I have an app that consists from engines exclusively, so there's no `app` folder at all. There's one non-isolated engine, which serves as an `app` folder, e.g. it holds the `application.html.erb` etc.

Right now every engine has its own `application.js` file, and the `application.html.erb` has `<%= stylesheet_link_tag engine_name %> where needed

But sprockets 4 requires me to have `manifest.js` located exactly at `app/assets/config`

rafaelfranca:
The manifest can be in any location that is part of the sprockets load path. `lib/assets/foo` is an example of this path. Is it not working to you?

dreyks:
It [raises](https://github.com/rails/sprockets-rails/blob/df5950017d7f2aa6fcbfa3949edfef85c35c28c7/lib/sprockets/railtie.rb#L105) `ManifestNeededError` if it's not there

rafaelfranca:
Oh, got it, yeah I think we should remove that check cc @schneems

schneems:
Thinking aloud

I don't see a benifit of making the location of the file customizable. While technically yes you can put that config in any file that will be auto loaded we need some way to know that the project is Sprockets 4 ready. I don't want to ease that requirement unless there's a good reason to do so. That's how I feel about Rails apps, but not engines.

I see a useful error message as, well useful. If we get rid of the error message we need something else to ease people into the upgrade process to Sprockets 4.

I do think we should allow assets in engines/gems to not have to change anything to support the new version. I guess I would prefer that those work out of the box by default and require fewer changes (if indeed they do work out of the box).

Maybe we could make the check for that file happen only after a precompile check failure?

dreyks:
If the main reason for a useful message is to ease the transition, maybe there could be some sort of a silencer. Like "we couldn't finds manifest.js in the correct location, please put it there. If you know what you're doing, set disable_manifest_strict_check = true to suppress this check"

move this check to "after-fail" is even cooler since it doesn't require any additional steps

schneems:
> If the main reason for a useful message is to ease the transition, maybe there could be some sort of a silencer

The silencer in this case is the presence of the file ;)

> moving this check to "after-fail" is even cooler since it doesn't require any additional steps

[...]

PR #446, opened 2019-10-10
Allow using a custom assets configuration manifest

kennyadsl:
Ref https://github.com/rails/sprockets-rails/issues/369

Using Sprockets 4, at the moment, the only allowed path for the newly required configuration manifest is the default one (`app/assets/config/manifest.js`) and there is no way to change it. 

This PR adds a way to change the default location by adding a new `config.assets.config_manifest` option, that could be set like this:

```ruby
config.assets.config_manifest = File.expand_path('my_custom_path/config/manifest.js', __dir__)
```

**Additional context**

In [Solidus](https://github.com/solidusio/solidus), we use a custom dummy app to perform our specs. We need to configure the path with custom locations for the configurations files we use. This PR allows us to set a custom configuration manifest path, otherwise, we would not be able to be compatible with Sprockets 4.  

Here are some relevant parts of our [DummyApp](https://github.com/solidusio/solidus/blob/674d0598c9b3f53a2434f2f0b02f5284481a0b98/core/lib/spree/testing_support/dummy_app.rb#L89-L96), I hope it helps to contextualize our issue.  

And this is the [failed build output](https://circleci.com/api/v1.1/project/github/solidusio/solidus/13827/output/106/0?file=true) (only first lines are relevant) now that sprockets 4 is used.

<hr>

This is my first contribution to this project so feel free to point out if I missed something important or how can I improve the PR. Thanks!

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -68,8 +68,8 @@ module Sprockets
     include Sprockets::Rails::Utils
 
     class ManifestNeededError < StandardError
-      def initialize
-        msg = "Expected to find a manifest file in `app/assets/config/manifest.js`\n" +
+      def initialize(manifest_path)
+        msg = "Expected to find a manifest file in `#{manifest_path}`\n" +
         "But did not, please create this file and use it to link any assets that need\n" +
         "to be rendered by your app:\n\n" +
         "Example:\n" +
@@ -99,11 +99,22 @@ module Sprockets
     config.assets.prefix      = "/assets"
     config.assets.manifest    = nil
     config.assets.quiet       = false
+    config.assets.config_manifest = nil
+
+    initializer :set_config_manifest_path do |app|
+      config_manifest = if app.config.assets.config_manifest.present?
+                          Pathname.new(app.config.assets.config_manifest)
+                        else
+                          ::Rails.root.join("app/assets/config/manifest.js")
+                        end
+      raise ManifestNeededError, config_manifest unless config_manifest.exist?
+
+      app.config.assets.config_manifest = config_manifest.to_s
+    end
 
     initializer :set_default_precompile do |app|
       if using_sprockets4?
-        raise ManifestNeededError unless ::Rails.root.join("app/assets/config/manifest.js").exist?
-        app.config.assets.precompile += %w( manifest.js )
+        app.config.assets.precompile += [app.config.assets.config_manifest]
       else
         app.config.assets.precompile += [LOOSE_APP_ASSETS, /(?:\/|\\|\A)application\.(css|js)$/]
       end
--- a/test/test_railtie.rb
+++ b/test/test_railtie.rb
@@ -12,6 +12,15 @@ ensure
   $stderr.reopen orig_stderr
 end
 
+def create_config_manifest(dir = "app/assets/config")
+  Dir.chdir(app.root) do
+    FileUtils.mkdir_p(dir)
+    File.open("#{ dir }/manifest.js", "w") do |f|
+      f << ""
+    end
+  end
+end
+
 class TestBoot < Minitest::Test
   include ActiveSupport::Testing::Isolation
 
@@ -39,13 +48,7 @@ class TestBoot < Minitest::Test
     @app.config.active_support.deprecation = :notify
     ActionView::Base # load ActionView
 
-    Dir.chdir(app.root) do
-      dir = "app/assets/config"
-      FileUtils.mkdir_p(dir)
-      File.open("#{ dir }/manifest.js", "w") do |f|
-        f << ""
-      end
-    end
+    create_config_manifest
   end
 
   def test_initialize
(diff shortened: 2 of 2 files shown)


ahorek:
here's an old discussion about this https://github.com/rails/sprockets-rails/issues/369 I really support to make this optional!

a few notes...
the error message has to fixed
https://github.com/rails/sprockets-rails/blob/49bf8022c8d3e1d7348b3fe2e0931b2e448f1f58/lib/sprockets/railtie.rb#L72
and there's a failing test
https://github.com/rails/sprockets-rails/blob/e9ca63edb6e658cdfcf8a35670c525b369c2ccca/test/test_railtie.rb#L298

kennyadsl:
Thanks, @ahorek! Not sure failure is related now.

lib/sprockets/railtie.rb
```diff
@@ -102,8 +106,13 @@ def configure(&block)
 
     initializer :set_default_precompile do |app|
       if using_sprockets4?
-        raise ManifestNeededError unless ::Rails.root.join("app/assets/config/manifest.js").exist?
-        app.config.assets.precompile += %w( manifest.js )
+        manifest_path = Pathname.new(
+          app.config.assets.manifest ||
+          ::Rails.root.join("app/assets/config/manifest.js")
+        )
+        raise ManifestNeededError unless manifest_path.exist?
```
ahorek:
```suggestion
        raise ManifestNeededError.new(manifest_path) unless manifest_path.exist?
```

lib/sprockets/railtie.rb
```diff
@@ -69,7 +69,11 @@ class Railtie < ::Rails::Railtie
 
     class ManifestNeededError < StandardError
       def initialize
-        msg = "Expected to find a manifest file in `app/assets/config/manifest.js`\n" +
+        manifest_path = Pathname.new(
```
ahorek: def initialize(manifest_path)
        msg = "Expected to find a manifest file in `#{manifest_path}`\n" +

test/test_railtie.rb
```diff
@@ -296,13 +296,21 @@ def test_sprockets_context_helper
   end
 
   def test_manifest_path
+    Dir.chdir(app.root) do
+      dir = "app/assets/config/foo"
+      FileUtils.mkdir_p(dir)
+      File.open("#{ dir }/bar.json", "w") do |f|
+        f << ""
+      end
+    end
```
elia:
```suggestion
    app.root.join("app/assets/config/foo/bar.json").tap { |path| path.dirname.mkpath }.write('')
```

elia (approved):
👍 FWIW, just one small suggestion for the spec

rafaelfranca:
We need a different config for this. The `assets.manifest` is the location of the output manifest. Maybe `config.assets.config_manifest`?

kennyadsl:
@rafaelfranca added a new config named `config_manifest` for this. Let me know if this works!

kennyadsl:
Another improvement (maybe with another PR) could be falling back to the Sprocket 3 behavior (`app.config.assets.precompile += [LOOSE_APP_ASSETS, /(?:\/|\\|\A)application\.(css|js)$/]`) if the configuration manifest does not exist. Do we really need to raise an exception if the configuration manifest is not present in the host application? Thoughts?

rafaelfranca:
Yes, we need to raise an exception. `LOOSE_APP_ASSETS` is a proc and procs are not supported anymore in the precompile option.

lib/sprockets/railtie.rb
```diff
@@ -99,11 +99,20 @@ def configure(&block)
     config.assets.prefix      = "/assets"
     config.assets.manifest    = nil
     config.assets.quiet       = false
+    config.assets.config_manifest = nil
+
+    initializer :set_config_manifest_path do |app|
+      config_manifest_path = Pathname.new(
+        app.config.assets.config_manifest ||
+        ::Rails.root.join("app/assets/config/manifest.js")
```
rafaelfranca: Let's set this default value on line 135.

kennyadsl: @rafaelfranca sorry, I'm not sure to get this request. Do you mean the whole `set_config_manifest_path ` initializer or just the default value of the preference? 

Setting the whole initializer at line 135 is not working since when used in the `set_default_precompile` initializer (before line 135) it is raising an exception since it's waiting for a string but it gets nil:

```
Error:
TestRailtie#test_task_precompile:
NoMethodError: undefined method `start_with?' for nil:NilClass
```

rafaelfranca: I mean setting `::Rails.root.join("app/assets/config/manifest.js")` (the default value when the config is not set) and removing this `||` from this intializer.

kennyadsl: I don't think it works this way. The Rails app has not yet been initialized when the code at that level is evaluated.

I'm doing this and I get:

```ruby
    # ...
    config.assets.manifest    = nil
    config.assets.quiet       = false
    config.assets.config_manifest = ::Rails.root.join("app/assets/config/manifest.js")

    initializer :set_config_manifest_path do |app|
      config_manifest_path = Pathname.new(app.config.assets.config_manifest)
      raise ManifestNeededError.new(config_manifest_path) unless config_manifest_path.exist?
      app.config.assets.config_manifest = config_manifest_path.to_s
    end
```

but `Rails.root` does not seem to be ready yet at that point; I get this exception:

```
undefined method `join' for nil:NilClass (NoMethodError)
```

I think that to use `::Rails.root` we need to call it from an initializer's context, right?

aldesantis:
@rafaelfranca any updates on this by any chance? It'd be really great to have it! 🙏

kennyadsl:
@rafaelfranca hey there, is there any thing I can do to move this forward?

kennyadsl:
@rafaelfranca friendly ping if you have time to point me in a good direction to move things forward here. Thanks!

fthobe:
@rafaelfranca any chance to reopen this?

AlbertGazizov:
Any chance to merge this?

PR #451, opened 2019-12-10
Remove deprecated alias_method_chain.

one-more-alex:


rafaelfranca:
Thank you for the pull request. This code is only used in version of rails were alias_method_chain is not deprecated, so the code is correct.

PR #454, opened 2019-12-27
Fix Ruby 2.7 deprecation in helper.rb

connorshea:
This is a sibling PR to https://github.com/rails/sprockets/pull/660. This needs to be changed to fix sprockets-rails causing a deprecation warning after the sprockets deprecation warning is fixed.

I've tested this on my Rails app and my test suite passes just fine with this change even if the sprockets change isn't present.

Oddly, this may actually be the only change that's necessary. I don't see any deprecation warning anymore with just this PR, regardless of whether the sprockets change is made or not.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -353,7 +353,7 @@ module Sprockets
 
         private
           def find_asset(path, options = {})
-            @env[path, options]
+            @env[path, **options]
           end
 
           def precompiled?(path)


nathanhinchey:
What version does this PR update? The latest available on rubygems.org is 4.0.0, which still has this issue.

renchap:
Would it be possible to have a new release of this gem so we do not need to install from git to get this Ruby 2.7 fix?
The last release was 3 years ago, it would be a good thing to properly release the recent changes to everybody (ping @rafaelfranca)

PR #460, opened 2020-04-21
Don't find all precompiled assets to check if one only is precompiled

mlisivka:
**Problem:**
After upgrading sprockets up to v4 we found that some of our pages load too slow (~20 sec) what was not in the previous version

**PR description:**
It happened because we didn't include the file extension explicitly to an asset path (ex. 'my_path/asset')
In this case, the gem tries to find all precompiled assets and then cache them but it looks like it does not work in my case, so the result I have is the gem tries to find assets on each request (~1250 assets).
I changed this behavior to check if one asset only


mlisivka:
I understood that this PR making development performance worse. So I close it

Issue #403, opened 2017-05-25
Manifest and Environment resolver inconsistencies


cainlevy:
The Environment resolver uses a forgiving search pattern that does not require assets such as images to specify their extension. The Manifest resolver does not. This means that given a `app/assets/images/foo.png`, syntax like `image_tag('foo')` will resolve fine locally and fail spectacularly when deployed to production.

Is this intended? It seems very error-prone.

teeceepee:
I encountered the same problem. My teammate insert `image_tag('bar')` in the index page, it worked in development environment, but crashed in production environment. That was terrible. I think we should make it consistent in any environment.

Issue #305, opened 2016-01-05
Image file extensions are now required (from v.3.0


virgoproz:
Until 2.3.3, it was ok to do image_tag('image')

From 3.0, to avoid getting the AssetsNotPrecompiled error, you need to mandatorily specify the file extension .... image_tag('image.png')

Again, not a big deal but a breaking change worth noting or fixing it is unwanted behavior accidentally added.

PR #461, opened 2020-05-14
A May of WTFs: Raise exception for assets declared without extension

lfv89:
Basically, it is possible to use `<%= image_tag('logo') %>` on development without any issues, but in production this always break because the file is not found without its extension. 

This has also been reported on: https://github.com/rails/sprockets-rails/issues/403, https://github.com/rails/sprockets-rails/issues/305 and https://discuss.rubyonrails.org/t/rails-6-0-2-image-tag-logo-works-in-development-but-crashes-in-production/74901/7

I tried to add the least amount of code to "fix" the problem. 

Let me know what you think.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -321,6 +321,8 @@ module Sprockets
         end
 
         def asset_path(path, digest, allow_non_precompiled = false)
+          raise ArgumentError, "asset #{path} should be declared with a filename extension" if File.extname(path).blank?
+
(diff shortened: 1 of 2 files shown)


lfv89:
@rafaelfranca any thoughts?

build is breaking but it doesn't seem related.

v2px:
Why was this closed? Still seems relevant to me and I just tripped over it in my Rails 6 deploy.

lfv89:
I closed it because sadly it doesn't look like there is any interest on fixing this ATM.

If I'm wrong I can gladly reopen it.

PR #467, opened 2020-09-19
Allow manifest.js to be placed anywhere.

### LINKED ISSUE #369: sprockets 4 manifest.js location

yubele:
# Description
This implementation ensures that you don't get an error if you place manifest.js anywhere (other than `app/assets/config/manifest.js`).

This fix may also help resolve #369.

# Problems to fix
After moving manifest.js, I get a [ManifestNeededError.](https://github.com/rails/sprockets-rails/blob/df5950017d7f2aa6fcbfa3949edfef85c35c28c7/lib/sprockets/railtie.rb#L105)

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -69,7 +69,7 @@ module Sprockets
 
     class ManifestNeededError < StandardError
       def initialize
-        msg = "Expected to find a manifest file in `app/assets/config/manifest.js`\n" +
+        msg = "Expected to find a manifest file in `manifest.js`\n" +
         "But did not, please create this file and use it to link any assets that need\n" +
         "to be rendered by your app:\n\n" +
         "Example:\n" +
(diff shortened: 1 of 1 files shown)


yubele:
In the test below, it is failed, but I ran it in the same procedure as travis, but it did not reproduce in my environment.

https://travis-ci.org/github/rails/sprockets-rails/jobs/728618169
https://travis-ci.org/github/rails/sprockets-rails/jobs/728618185

PR #468, opened 2020-09-25
List manifest locations when asset is not found

schneems:
When using an asset manifest file `config.assets.compile = false` the developer might not know that a `assets:precompile` manifest file is being used, or they might not know which file (if there are multiple). To help debug this situation we can detect which asset resolution strategies contain a "file" and output the location of that file:

```
The asset "does_not_exist.js" is not present in the asset pipeline.
Checked in manifest file: /Users/rschneeman/Documents/projects/sprockets-rails/tmp/app/public/assets/.sprockets-manifest-7bdd432004acd447a520e289399bcf62.json
```

I lost about 30 minutes debugging an error I was seeing locally while running with RAILS_ENV=production. It was telling me a file "was not present in the asset pipeline" which I took to mean it wasn't in `app/assets` or maybe the config declaration was wrong. Everything looked right, it wasn't until much later I realized that I had somehow gotten an old manifest file generated from an older `assets:precompile`run so it was checking that file to see if the asset was "present in the pipeline". 

By adding the location of the manifest file to the output, it's more likely that I would have realized that sprockets was only checking that file and that the solution was to re-generate it.

### DIFF
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -80,7 +80,11 @@ module Sprockets
         if asset_path = resolve_asset_path(path, debug)
           File.join(assets_prefix || "/", legacy_debug_path(asset_path, debug))
         else
-          message =  "The asset #{ path.inspect } is not present in the asset pipeline.\n"
+          message = String.new("The asset #{ path.inspect } is not present in the asset pipeline.\n")
+          asset_resolver_strategies.each do |strat|
+            message << "Checked in manifest file: #{strat.filename}\n" if strat.filename
+          end
+
           raise AssetNotFound, message unless unknown_asset_fallback
 
           if respond_to?(:public_compute_asset_path)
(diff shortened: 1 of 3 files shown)


schneems:
These tests are not good. We need to move off of travis. Kill support for older rubies. Kill support for older Rails versions (sorry, if Rails core doesn't support it anymore, I don't want to support it either).

Ideally, I would LOVE to kill this gem too and put it directly in sprockets, but I don't know if it's worth it. I believe it's not there because Josh didn't want to be coupled to Rails at all. But as a result, this gem is much more complicated as it has to support a matrix of Rails and Sprocket versions.

skatkov:
@schneems you want to migrate off travis, what exactly do you want to migrate to? Github actions?

schneems:
I'm not super happy with GitHub due to their involvement with ICE so I try to not use actions if I can avoid it. I have the most experience with CircleCI. It looks like Rails switched over to build kite. I'm asking in the basecamp chat to see if there's a policy of sorts.

skatkov:
Will be happy to give you hand with this as well, once you'll arrive at final decision. 

Like you, I have way more experience with CircleCI, but buildkite is doable too.

PR #469, opened 2020-09-27
Travis-ci: remove unsupported rubies and rails versions

skatkov:
Removing some unsupported stuff from travis ci matrix - rubies and rails.

Based on a comment shared by @schneems in https://github.com/rails/sprockets-rails/pull/468#issuecomment-699177569


rafaelfranca:
> I don't think we can kill off 5.1 support just yet until Rails 6.1 comes out.

Yes, we can. Only Rails 5.2 and 6.0 are currently supported.

I don't think we need to release a major version of sprockets-rails just to drop support to old Rails version. We only need to update the sprockets-rails.gemspec to not allow it being installed in any Rails < 5.2.

skatkov:
Agree with @rafaelfranca. Not sure why we can't drop 5.1, since only 5.2 and 6.0 are supported now.

I also updated gemspec.

schneems:
> Only Rails 5.2 and 6.0 are currently supported.

Right, sorry. I got the Rails support policy (2 releases) and Ruby (3 releases) mixed up in my head. Thanks for the clarification. I've been away from github.com/rails for too long 😅

Thanks for the work here!

PR #475, opened 2021-06-17
Move to github actions

hahmed:
## What

Move to GitHub actions and drop TravisCI.

- `Gemfile` has been update to point to `rails:main` because `rails:master` no longer exists
- I have locked rack to a [specific commit](https://github.com/rails/sprockets-rails/pull/475/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR7) because the latest Rack version breaks on rails see https://github.com/rack/rack/pull/1751
- The [method signature](https://github.com/rails/sprockets-rails/pull/475/files#diff-18ee2da213613df40be5a270090e93d8af492610df02ba40220c18e4a8a19cc6R428) has been updated for `ActionView::Base.new`, we now pass in `{}, null` for the locals and controller respectively.
- [The attribute](https://github.com/rails/sprockets-rails/pull/475/files#diff-ba37813ca277c227a74a372479b7b05b7f3ff085d890ab708f80d62573efdb7aR11) `media="screen"` which has now been removed see https://github.com/rails/rails/pull/41215 is needed for the tests to pass for Rails versions < 7. Rather than instantiation a rails app and setting the config `Rails.application.config.action_view.stylesheet_media_default=true`, this seems like the easier fix.

## Anything else

I have split into 2 commits, 1 for the build (which is currently failing on travis ci) and the other for the test fixes, but let me know if you want me to squash into 1 commit.

The GitHub actions are not being run here, but you can see them all green here https://github.com/hahmed/sprockets-rails/actions/runs/952618642 -- not sure why I can see the actions workflow on my fork but not upstream, but @byroot mentioned if we merged this PR we would prob see the actions here too.

Should we consider updating the target ruby version in a follow up PR, perhaps ruby 2.7?


hahmed:
fyi... https://github.com/rack/rack/pull/1755 reverts `Flatten Rack::Request::Env into Request` we could target the maser branch once that get's merged.

PR #479, opened 2021-11-12
Process source mapping URLs be set by transpilers

dhh:
If you prepare a JavaScript file by a bundler that adds a sourceMappingURL comment line as the last in the file, Sprockets will append a semicolon to this comment, thus breaking the URL reference.

The sourceMappingURL will also reference a file that has not been digested yet.

This PR fixes both issues.


dhh:
Reference https://github.com/rails/sprockets/issues/702 for history about this semicolon appending business.

PR #482, opened 2021-11-19
Strip away the relative path

### LINKED ISSUE #478: Incorrect relative URL in css
We have updated sprockets-rails to 3.3.0 and now relative urls are wrong. With his line in css:

`src: url(./fontawesome/fa-brands-400.eot);`

the url is transformed to:

`src: url(/./fontawesome/fa-brands-400.eot);`

The first slash is incorrect because we have the font files in a path relative to the css, not in the root

jcoyne:
Fixes #478

Previously having this css:
```
src: url(./fontawesome/fa-brands-400.eot);
```

would be transformed into:
```
src: url(/./fontawesome/fa-brands-400.eot);
```

This strips off the unnecessary `./` so it will be:
```
src: url(/fontawesome/fa-brands-400.eot);
```

### DIFF
--- a/lib/sprockets/rails/asset_url_processor.rb
+++ b/lib/sprockets/rails/asset_url_processor.rb
@@ -2,11 +2,13 @@ module Sprockets
   module Rails
     # Rewrites urls in CSS files with the digested paths
     class AssetUrlProcessor
-      REGEX = /url\(\s*["']?(?!(?:\#|data|http))([^"'\s)]+)\s*["']?\)/
-
+      REGEX = /url\(\s*["']?(?!(?:\#|data|http))(?<relativeToCurrentDir>.\/)?(?<path>[^"'\s)]+)\s*["']?\)/
       def self.call(input)
         context = input[:environment].context_class.new(input)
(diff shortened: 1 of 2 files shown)


jcoyne:
@chuchiperriman does this solve your issue?  It wasn't entirely clear to me the situation you had given the description for #478

dhh:
I think the problem was also with fully relative paths, like: `src: url(fontawesome/fa-brands-400.eot)`. What actually is the right thing for us to do then? I think that too has to be converted to `/fontawesome/fa-brands-400.eot`.

jcoyne:
@dhh I have the same question. I added a test for the current behavior for that case here: https://github.com/rails/sprockets-rails/pull/482/files#diff-8226e061e1adf779cb31d000395d5c814bc94696f12b25fa4ac7cb8acb9ee555R58-R63 I don't know if that's the right behavior though.

dhh:
Other option is to truly let it be relative. But don't know if we have all the info for the sprockets processor to do that?

dhh:
Anyway, this at least fixes the direct breakage.

Issue #481, opened 2021-11-16
Infinite recursion in SourcemappingUrlProcessor


tomhughes:
We have an asset in our project that appears to be causing an infinite recursion in the new `SourcemappingUrlProcessor` with a stack that looks like:

```
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-rails-3.4.0/lib/sprockets/rails/sourcemapping_url_processor.rb:15:in `block in call'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-rails-3.4.0/lib/sprockets/rails/sourcemapping_url_processor.rb:12:in `gsub'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-rails-3.4.0/lib/sprockets/rails/sourcemapping_url_processor.rb:12:in `call'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:84:in `call_processor'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:66:in `block in call_processors'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:65:in `reverse_each'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:65:in `call_processors'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:22:in `block in <class:CompositeProcessor>'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:33:in `call'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:84:in `call_processor'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:66:in `block in call_processors'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:65:in `reverse_each'
/usr/local/share/gems/ruby/3.0.0/gems/sprockets-4.0.2/lib/sprockets/processor_utils.rb:65:in `call_processors'
[...]

PR #483, opened 2021-11-22
Only process sourceMappingURLs for files, not paths

dhh:
This will prevent the recursion issue noticed in #481.

### DIFF
--- a/test/test_sourcemapping_url_processor.rb
+++ b/test/test_sourcemapping_url_processor.rb
@@ -24,6 +24,16 @@ class TestSourceMappingUrlProcessor < Minitest::Test
     assert_equal({ data: "var mapped;\n//# sourceMappingURL=mapped-HEXGOESHERE.js.map\n//!\n" }, output)
   end
 
+  def test_prevent_recursion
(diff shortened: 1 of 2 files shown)


test/test_sourcemapping_url_processor.rb
```diff
@@ -24,6 +24,16 @@ def asset_path(path, options = {})
     assert_equal({ data: "var mapped;\n//# sourceMappingURL=mapped-HEXGOESHERE.js.map\n//!\n" }, output)
   end
 
+  def test_prevent_recursion
+    input = { environment: @env, data: "var mapped;\n//# sourceMappingURL=/assets/mapped.js.map", filename: 'mapped.js', metadata: {} }
```
rafaelfranca, line 28: Source maps inside directories are not supported?

dhh: Currently not.

dhh:
This wasn't it.

test/test_sourcemapping_url_processor.rb
dhh: Managed to add support in this new PR I'm working on.

PR #489, opened 2021-12-06
Add `assets.resolve_assets_in_css_urls` configuration option to allow disabling `AssetUrlProcessor`

rmacklin:
As reported in https://github.com/rails/sprockets-rails/issues/478 / https://github.com/rails/sprockets-rails/issues/486, the introduction of the `AssetUrlProcessor` in https://github.com/rails/sprockets-rails/pull/476 broke some folks' CSS assets when they upgraded beyond sprockets-rails v3.2.2, but it was not released in a new major version.

~To remedy this, let's make `AssetUrlProcessor` be opt-in for now with the plan to change it to opt-out in the next major version. In this PR, I'm proposing the addition of a new `assets.rewrite_css_urls` configuration option to control whether or not `AssetUrlProcessor` is registered. We can default it to `false` in sprockets-rails v3.x and change the default to `true` in v4.0.~
Per https://github.com/rails/sprockets-rails/pull/489#issuecomment-988582781, we'll roll forward with defaulting `AssetUrlProcessor` to enabled, but it can be disabled by setting a new `assets.resolve_assets_in_css_urls` configuration option to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ config.assets.configure do |env|
 end
 ```
 
+**`config.assets.resolve_assets_in_css_urls`**
+
+When this option is enabled, sprockets-rails will register a CSS postprocessor to resolve assets referenced in [`url()`](https://developer.mozilla.org/en-US/docs/Web/CSS/url()) function calls and replace them with the digested paths. Defaults to `true`.
+
 **`config.assets.resolve_with`**
 
 A list of `:environment` and `:manifest` symbols that defines the order that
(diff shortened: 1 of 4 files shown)


dhh:
I think this is a good option, but I don't see reverting what's now been released and out in the wild. If the 3.4.x series breaks things for you, you can set this option to avoid having the processor applied. But otherwise it's applied.

rmacklin:
Ok, I dropped that change and added tests for the new configuration option.

dhh:
This is looking good. We should document the option as well.

Also, I think going with `resolve_assets_in_css_urls` might be a more descriptive name for the configuration.

rmacklin:
Ok, renamed the option to `resolve_assets_in_css_urls` and documented it in the README.

rmacklin:
Ok, renamed the option to `resolve_assets_in_css_urls` and documented it in the README.

PR #510, opened 2022-07-13
Remove legacy conditions

skipkayhil:
### Summary

These are no longer necessary to due higher Ruby and Rails requirements


lib/sprockets/railtie.rb
```diff
@@ -55,13 +55,10 @@ def precompiled_assets(clear_cache = false)
...
-      initializer :append_assets_path, :group => :all do |app|
-        app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
-        app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
-        app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)
-      end
+    initializer :append_assets_path, :group => :all do |app|
+      app.config.assets.paths.unshift(*paths["vendor/assets"].existent_directories)
+      app.config.assets.paths.unshift(*paths["lib/assets"].existent_directories)
+      app.config.assets.paths.unshift(*paths["app/assets"].existent_directories)
     end
```
skipkayhil, line 62: Could `Sprockets::Railtie` be upgraded to a `Rails::Engine` so that we don't have to reopen the `Rails::Engine` class? (`Rails::Railtie` doesn't have access to the `paths` used here)

lib/sprockets/rails/route_wrapper.rb
```diff
@@ -1,23 +1,13 @@
...
       end
-
-      def self.included(klass)
-        klass.class_eval do
-          def internal_with_sprockets?
-            internal_without_sprockets? || internal_assets_path?
-          end
-          alias_method_chain :internal?, :sprockets
-        end
-      end
```
p8, line 20: So this can be removed because we no longer call `include Sprockets::Rails::RouteWrapper` ?

skipkayhil: Yep, exactly

lib/sprockets/railtie.rb
p8: I think we can:
> When we drop support to Rails 4 we can change it to instead of reopening Rails::Engine just define a Sprockets::Rails::Engine instead of a Railtie.

https://github.com/rails/sprockets-rails/pull/215#discussion_r23323619

rafaelfranca: We can't do that. This needs to be an initializer defined in all `Rails::Engine` subclasses, not only in the sprockets Engine. But we can do the same as https://github.com/rails/propshaft/pull/43.

skipkayhil: Hmm, can you elaborate on the difference? I thought https://github.com/rails/propshaft/commit/66ce0f889377b81a28b1dd1fbe245868a047d46f didn't work because `Propshaft::Railtie` isn't a full engine.

rafaelfranca: The difference is that what this code is doing is changing all engines (every single one of them) and defining an initializer insider then. If we just move that initializer to the SprocketsRails Engine, only it would be affected.

skipkayhil: Ah, it just clicked for me: the initializer is appending the `Rails::Engine` subclasses' `paths` to `config.assets.paths`. That makes complete sense now.

PR #511, opened 2022-07-15
In url() prevent relative path from being converted to root absolute path

Sega100500:
Prevent non-digest  (not assets pipeline) relative path from being converted to root absolute path. Prevent convert url() in attached third-party css-styles libraries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 Gemfile*.lock
 tmp/
 *.gem
+.idea
--- a/lib/sprockets/rails/asset_url_processor.rb
+++ b/lib/sprockets/rails/asset_url_processor.rb
@@ -7,7 +7,14 @@ module Sprockets
         context = input[:environment].context_class.new(input)
         data    = input[:data].gsub(REGEX) do |_match|
           path = Regexp.last_match[:path]
-          "url(#{context.asset_path(path)})"
+          # original: "url(#{context.asset_path(path)})"
+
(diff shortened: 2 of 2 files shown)


shuyahonda:
@Sega100500 
 I'm using ckeditor gem and meet the same problem.
This fix will solve the problem.

huda-kh:
> @Sega100500 I'm using ckeditor gem and meet the same problem. This fix will solve the problem.

Same here.. what have you done while waiting for this fix?

Sega100500:
@huda-kh 
I suggested a very simple solution for fix this bug - just 2 lines of code. My fix failed the some tests and was successfully ignored.
It is **not difficult for me to manually fix the gem code after installation**.

If Rails developers do not care about their product (and included to it gems) - they continue to ignore this bug.

huda-kh:
This fixed it for me. https://github.com/galetahub/ckeditor/issues/919#issuecomment-1372890417On 16 Aug 2024 7:48 pm, Sega ***@***.***> wrote:
@huda-kh
I suggested a very simple solution for fix this bug - just 2 lines of code. My fix failed the some tests and was successfully ignored.
It is not difficult for me to manually fix the gem code after installation.
If Rails developers do not care about their product (and included to it gems) - they continue to ignore this bug.

Sega100500:
@huda-kh 
Yes, agree - using ```config.assets.resolve_assets_in_css_urls``` is also very good solution.

PR #513, opened 2022-12-05
Update README.md to clarify url(...) references in CSS in processed CSS

jasonfb:
@jrochkind

adds language to README: "It no longer necessary to use eg asset-url() helper (which was introduced by sass-rails). All `url(...)` references in CSS are just transparently processed to use appropriate digested filenames automatically now."


jasonfb:
https://github.com/rails/sprockets-rails/pull/476

januszm:
It'd be also great to include this information in https://github.com/rails/dartsass-rails (especially the part about changing `image-url` to just `url`)

PR #517, opened 2023-04-06
Use a dedicated ActiveSupport::Deprecation

etiennebarrie:
Rails 7.1 will deprecate using the singleton ActiveSupport::Deprecation instance (https://github.com/rails/rails/pull/47354). This defines one for the gem and uses it.

It's also added to the application's set of deprecators, allowing it to be configured along all the other deprecators.

cc @rafaelfranca

### DIFF
--- a/lib/sprockets/railtie.rb
+++ b/lib/sprockets/railtie.rb
@@ -125,6 +125,10 @@ module Sprockets
       Sprockets.register_postprocessor "application/javascript", ::Sprockets::Rails::SourcemappingUrlProcessor
     end
 
+    initializer "sprockets-rails.deprecator" do |app|
+      app.deprecators[:sprockets_rails] = Sprockets::Rails.deprecator if app.respond_to?(:deprecators)
+    end
+
     config.assets.version     = ""
     config.assets.debug       = false
     config.assets.compile     = true
--- a/lib/sprockets/rails.rb
+++ b/lib/sprockets/rails.rb
@@ -1,4 +1,13 @@
 require 'sprockets/rails/version'
+require 'active_support'
 if defined? Rails::Railtie
   require 'sprockets/railtie'
+end
+
+module Sprockets
+  module Rails
+    def self.deprecator
+      @deprecator ||= ActiveSupport::Deprecation.new("4.0", "Sprockets::Rails")
+    end
+  end
 end
--- a/lib/sprockets/rails/helper.rb
+++ b/lib/sprockets/rails/helper.rb
@@ -89,7 +89,7 @@ module Sprockets
             message << "To bypass the asset pipeline and preserve this behavior,\n"
             message << "use the `skip_pipeline: true` option.\n"
 
-            ActiveSupport::Deprecation.warn(message, caller_locations)
+            Sprockets::Rails.deprecator.warn(message, caller_locations)
           end
           super
         end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -2,6 +2,7 @@ require 'minitest/autorun'
 
 require 'action_view'
 require 'sprockets'
+require 'sprockets/rails'
 require 'sprockets/rails/context'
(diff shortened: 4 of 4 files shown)


lib/sprockets/railtie.rb
```diff
@@ -125,6 +125,10 @@ def configure(&block)
       Sprockets.register_postprocessor "application/javascript", ::Sprockets::Rails::SourcemappingUrlProcessor
     end
 
+    initializer :deprecator do |app|
```
skipkayhil, line 128:
```suggestion
    initializer "sprockets-rails.deprecator" do |app|
```

While the other initializers here don't do this (and probably never will due to backwards incompatibility), I think any new initializers added should be namespaced so that they don't conflict with any other gems.

(Duplicate initializer names has been a problem before: https://github.com/rails/rails/pull/43261#issuecomment-930336800)

skipkayhil (commented):
One comment, otherwise looks good to me!

lib/sprockets/railtie.rb
etiennebarrie: Hum there's already a namespace in the sense that all initializers are defined within a Railtie.
```
$ bin/rails initializers | grep deprecator
Sprockets::Railtie.deprecator
```
(Rails 7.0 app running with this branch)

It feels like the issue you're mentioning is more a bug/misfeature in `Railtie::Initializable` but OK.

I see all the other deprecators have a namespace in their initializer's name so that makes sense.

ghiculescu:
Could we get this included in a release please?

klyonrad:
This change causes an exception when we try to initialize our Rails 7.1 app

```
gems/sprockets-rails-3.5.0/lib/sprockets/railtie.rb:129:in `block in <class:Railtie>': undefined method `deprecator' for Sprockets::Rails:Module (NoMethodError)

      app.deprecators[:sprockets_rails] = Sprockets::Rails.deprecator if app.respond_to?(:deprecators)
```

what could we do to fix that?

etiennebarrie:
What do you mean by initialize your app?
I can't reproduce with a brand new Rails 7.1 app with sprockets-rails 3.5.0, I'm able to boot the application and precompile assets.

Can you share a backtrace or a repro script?

fa11enangel:
> This change causes an exception when we try to initialize our Rails 7.1 app
> 
> ```
> gems/sprockets-rails-3.5.0/lib/sprockets/railtie.rb:129:in `block in <class:Railtie>': undefined method `deprecator' for Sprockets::Rails:Module (NoMethodError)
> 
>       app.deprecators[:sprockets_rails] = Sprockets::Rails.deprecator if app.respond_to?(:deprecators)
> ```
> 
> what could we do to fix that?

Maybe it is solved by newer version, see https://github.com/rails/sprockets-rails/pull/525

etiennebarrie:
Ok sorry about that (#524), it seems Rails itself used to require `sprockets/railtie` directly before 7.0 in `config/application.rb` [via `rails/all`](https://github.com/rails/rails/commit/fb1ab3460a676ce7def0819c2e92289ef2dcbe3b#diff-22357f8be507e1f3f2366d30c196086a10d34b33c983a2f1be36750bb473bd4cL18) or more importantly [directly if not all frameworks](https://github.com/rails/rails/commit/fb1ab3460a676ce7def0819c2e92289ef2dcbe3b#diff-907b37333a84c0b6faee93e22de9cca471e2bf8c66c8e13568f0bf508843186fL18) were loaded.

So there can be a lot of apps that directly require `sprockets/railtie` without loading `sprockets-rails`.

Similarly if you use only `sassc-rails` in your Gemfile, only the railtie is loaded:
https://github.com/sass/sassc-rails/blob/v2.1.2/lib/sassc/rails/railtie.rb#L4

I've opened https://github.com/rails/sprockets-rails/pull/527 to make sure this doesn't happen again.

PR #522, opened 2024-02-21
Handle Inline Sourcemaps

nicklozon:
# Description
Sprockets gem appends semicolons to each js asset which corrupts `//# sourceMappingURL=` magic comments. `sprockets-rails` fixes this by appending `\n//!\n` to the asset so the `sourceMappingURL` is not the last line and in turn not corrupted by the semicolon. This only applies to external sourcemap files.

This PR improves `SourcemappingUrlProcessor` to also add `\n//!\n` after `sourceMappingURL` comments when they contain inline sourcemaps.

### DIFF
--- a/lib/sprockets/rails/sourcemapping_url_processor.rb
+++ b/lib/sprockets/rails/sourcemapping_url_processor.rb
@@ -2,20 +2,24 @@ module Sprockets
   module Rails
     # Rewrites source mapping urls with the digested paths and protect against semicolon appending with a dummy comment line
     class SourcemappingUrlProcessor
-      REGEX = /\/\/# sourceMappingURL=(.*\.map)/
+      FILE_REGEX = /\/\/# sourceMappingURL=(.*\.map)/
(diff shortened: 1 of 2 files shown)


lib/sprockets/rails/sourcemapping_url_processor.rb
```diff
@@ -2,21 +2,25 @@ module Sprockets
   module Rails
     # Rewrites source mapping urls with the digested paths and protect against semicolon appending with a dummy comment line
     class SourcemappingUrlProcessor
-      REGEX = /\/\/# sourceMappingURL=(.*\.map)/
+      FILE_REGEX = /\/\/# sourceMappingURL=(.*\.map)/
+      INLINE_REGEX = /\/\/# sourceMappingURL=(data:.*)$/
```
nicklozon, line 6: Typically inline sourcemaps will look something like this: `sourceMappingURL=data:application/json;base64,abc123=` where `abc123` is encoded data. The regex assumes anything starting with data: is an inline sourcemap, which I fell is a safe assumption from my experience.

PR #543, opened 2024-10-10
Fix asset url processor regex for where subdirectory starts with 'data'

mikevoets:
Our precompilation using sprockets-rails was skipping over processing asset urls that refer to assets inside `app/assets/images/data`.

This caused it to not resolve assets referenced in with for example `url('data/foo.png')` function calls and replace them with the digested paths.

This was caused by the regular expression in `Sprockets::Rails::AssetUrlProcessor::REGEX`, which wrongfully considered assets in `data/` directories to be `data:` protocol assets.

This PR fixes the regex to not skip over url() function calls referring to assets in `data/` directories and only skip when the `data` or `http(s)` protocol in the url is followed by a colon.


mikevoets:
Closing because we have a workaround and we will be moving on from sprockets soon.

PR #546, opened 2025-02-14
Warn instead of raising on missing `manifest.js`

mamhoff:
With a pristine (`rails new`) Rails 8.0 app, the `app/assets/config/manifest.js` file is no longer automatically generated. For engines still using sprockets, that is not a huge issue, because they [have and add their own manifest files](https://github.com/solidusio/solidus/blob/main/backend/config/initializers/assets.rb) for their assets to the list of precompiled assets.

However, scripts like the [Solidus bin/sandbox script](https://github.com/solidusio/solidus/blob/main/bin/sandbox#L44-L99) first run `rails new`, then add the needed engines to the sandbox's `Gemfile`, then run `rails db:migrate` and their respective install generators. With Rails 8, these scripts now fail before anything is possible in Ruby, because the initializer of `Sprockets::Railtie` will raise an error on startup.

This commit changes the behavior of the Railtie to issue a deprecation warning instead of raising an Exception.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,35 +9,43 @@ jobs:
         include:
           - ruby: 2.5
             gemfile: "gemfiles/Gemfile.rails-6.1-sprockets-3"
+            bundler: "2.3.0"
           - ruby: 2.5
             gemfile: "gemfiles/Gemfile.rails-6.1-sprockets-4"
-
+            bundler: "2.3.0"
           - ruby: 2.7
             gemfile: "gemfiles/Gemfile.rails-7.0-sprockets-3"
+            bundler: "2.4.8"
           - ruby: 2.7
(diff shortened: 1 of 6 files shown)


mamhoff:
Rebased on #547 to get the CI green.

tvdeyen (commented):
Thanks. This will help every Rails 8 app to even boot when a gem depends on `sprockets-rails`. Please merge this 🙏🏻

tvdeyen:
@rafaelfranca sorry for the ping, but someone please needs to merge this as this breaks all Rails apps still using sprockets. 

Try this

```
gem install rails -N
rails new my_app
cd my_app
bundle add sprockets-rails
bin/rails c
💣 💥
```

Noah-Silvera:
Just ran into this problem again when creating a new rails app with Solidus

If anyone has time to take another look at at this PR, I think it will save other folks confusion as time goes on. 🙏

PR #547, opened 2025-02-15
CI: Specify Bundler version

mamhoff:
Bundler `latest` has stopped working with older Rubies. We need to specify the Bundler version for the older Rubies to still build.

Also adds the `concurrent-ruby` limit to `< 1.3.5` for Rails 6.1 and 7.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,35 +9,43 @@ jobs:
         include:
           - ruby: 2.5
             gemfile: "gemfiles/Gemfile.rails-6.1-sprockets-3"
+            bundler: "2.3.0"
           - ruby: 2.5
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -49,6 +57,8 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          cache-version: ${{ matrix.ruby }}-${{ matrix.ruby }}-${{ matrix.gemfile }}
```
tvdeyen, line 60: You probably wanted?

```diff
-${{ matrix.ruby }}-${{ matrix.ruby }}-${{ matrix.gemfile }}
+${{ matrix.ruby }}-${{ matrix.bundler }}-${{ matrix.gemfile }}
```

